PR #39, opened 2020-11-06
Vertx Webclient and kafka use cases

pjgg:
Vert.x Mutiny webClient exploratory test.
* Vert.x WebClient
* Quarkus Resteasy Mutiny / Jsonb
* Quarkus configuration converters
* Exception mapper
* Automating testing

Vert.x Kafka exploratory test.
* Kafka producer
* Kafka consumer
* AVRO serializers/deserializers
* Application property profiles
* Deploy Verticles
* Automating testing

PR #40, opened 2020-11-06
Added gRPC with Kafka and Infinispan example

Sgitario:
These changes contain:
- New section for gRPC examples from 700 to 799. 
- A new module called Added gRPC with Kafka and Infinispan example.
- A common module to have all the common utilities like the test containers.


700-quarkus-infinispan-grpc-kafka/src/main/java/io/quarkus/qe/books/Book.java
```diff
@@ -0,0 +1,71 @@
...
+
+public class Book {
+
+    private String title;
+    private String description;
+    private int publicationYear;
+
+    public Book() {
+
+    }
```
Sgitario, line 18: This empty constructor is necessary to use this class in the rest endpoint (not a good practice though).

rsvoboda:
@Sgitario did you try native mode ?

Sgitario:
> @Sgitario did you try native mode ?

Yes, using:

```
mvn clean install -Dnative -Dquarkus.native.container-build=true -Dquarkus.native.native-image-xmx=4g
```

common-test-utils/pom.xml
```diff
@@ -0,0 +1,28 @@
...
+
+    <dependencies>        
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>kafka</artifactId>
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-test-common</artifactId>
```
rsvoboda, line 24: This makes this repo tightly coupled with infrastructure module in main Quarkus repo.
I do not like it too much. What are the benefits of this approach?

common-test-utils/pom.xml
```diff
@@ -0,0 +1,28 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.quarkus.qe</groupId>
+        <artifactId>beefy-scenarios</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>common-test-utils</artifactId>
```
rsvoboda, line 14: This module breaks the independence of individual modules. Atm this helps to share one InfinispanTestResource in 2 modules.

I personally prefer to keep things loosely coupled with small degree of duplicates if needed.

@mjurc what's your view?

pom.xml
```diff
@@ -46,6 +50,19 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+
```
rsvoboda, line 53: Are these new dependencies used by all the modules or just subset of modules?

common-test-utils/pom.xml
Sgitario: This is to have the QuarkusTestResource annotation into this module. We can use the quarkus junit5 dependency instead.

pom.xml
Sgitario: Just a subset. This is just to have the ability of sharing common logic across modules.

common-test-utils/pom.xml
rsvoboda: I prefer to use things intended for end users, quarkus-test-common is not mentioned in quarkus docs / guides at all.

It's a helper in main Quarkus repo and I would keep it that way. Apps in this repo should try to avoid such helpers as they can pull in a lot of stuff and can hide something that way.

pom.xml
```diff
@@ -46,6 +50,19 @@
...
+        <groupId>io.quarkus.qe</groupId>
+        <artifactId>common-test-utils</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      
+      <dependency>
+        <groupId>org.testcontainers</groupId>
+        <artifactId>kafka</artifactId>
+        <version>${testcontainers.version}</version>
+        <scope>test</scope>
```
kshpak, line 64: Testcontainers are not used across all modules, I would prefer to import it in individual module that requires it

common-test-utils/pom.xml
kshpak: @rsvoboda I agree

pom.xml
rsvoboda: I'm not really in favor of sharing common stuff across modules.

At this stage I still hesitate to introduce such module. I still believe in having all the things inside the module for easy extraction + keeping just the required deps per the module.

My experience with shared / common modules is not great, people tend to drop more and more things there and over time it becomes hell to maintain it and keep stuff in reasonable boundaries.

pom.xml
Sgitario: The module is declared within the dependency management, so this dependency won't be automatically added across all the modules. 
If one module wants to use it, the module needs to add it as a dependency.

common-test-utils/pom.xml
Sgitario: Moving this discussion into the thread: https://github.com/quarkus-qe/beefy-scenarios/pull/40#discussion_r519770567

pom.xml
Sgitario: I see your point. We could remove this common utils and have an infinispan module for the test resource only, but taking into account that there are only two modules that would use it, it does not bring up a lot of value. 
Will remove this module and have the infinispan test resource duplicated.

common-test-utils/pom.xml
Sgitario: Will remove this module

pom.xml
rsvoboda: One more thing, people need to run `mvn test -pl 700-quarkus-infinispan-grpc-kafka,common-test-utils` instead of `mvn test -pl 700-quarkus-infinispan-grpc-kafka` and it is easy to forget about it / use old common-test-utils installed in local repo

README.md
```diff
@@ -15,6 +15,7 @@ Beefy scenarios for Quarkus
 | 400 - 499     | QuarkConf | Quarkus scenarios presented at conferences |
 | 500 - 599     | QuarkUser | Quarkus scenarios adopted by end users |
 | 600 - 699     | QuarkSpring | Quarkus scenarios with Spring experience |
+| 700 - 799     | QuarkGrpc | Quarkus scenarios with gRPC experience |
```
rsvoboda, line 18: I think this should bee in `000 - 099 - Quarkus only scenarios` set

We might want to rename it to `Quarkus native scenarios`

700-quarkus-infinispan-grpc-kafka/src/main/proto/helloworld.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
```
rsvoboda, line 1: Is this .proto file included intentionally?

README.md
Sgitario: Ok

700-quarkus-infinispan-grpc-kafka/src/main/proto/helloworld.proto
Sgitario: Yes, this is to test the gRPC service.

rsvoboda: please make a note in readme about this file, not everybody is aware of protobuf stuff / have deep grpc insight

rsvoboda (changes requested):
@Sgitario please make the module more standalone, drop deps on Quarkus test common artifact and please also drop the common module to keep the modules independent. Please ping me if you want to discuss more in depth.

From functional perspective I haven't review the code closely enough, will do that once the above get reflected.

pom.xml
Sgitario: Removed

README.md
Sgitario: Changed

700-quarkus-infinispan-grpc-kafka/src/main/proto/helloworld.proto
Sgitario: Added

Sgitario:
> @Sgitario please make the module more standalone, drop deps on Quarkus test common artifact and please also drop the common module to keep the modules independent. Please ping me if you want to discuss more in depth.
> 
> From functional perspective I haven't review the code closely enough, will do that once the above get reflected.

I think I've already addressed all your requests. Can you review it again? Thanks.

rsvoboda:
I have no further objections to the code and functionality, I checked native too. Good job @Sgitario.

I'm not approving this yet as I want to merge https://github.com/quarkus-qe/beefy-scenarios/pull/39 first so you can rebase + fix conflicts + adjust your PR a bit - no explicit testcontainers version definition as it will be defined in dependency management in root pom.

rsvoboda:
@Sgitario https://github.com/quarkus-qe/beefy-scenarios/pull/39 is merged, please adjust this PR to use dependency management defined testcontainers

Sgitario:
> @Sgitario #39 is merged, please adjust this PR to use dependency management defined testcontainers

Done. I'm now using the testcontainers dependencies provided by the parent.

rsvoboda (approved):

mjurc (approved):

PR #42, opened 2020-11-10
Added opentracing, resteasy, mutiny example

Sgitario:
Module that test whether the OpenTracing data is transmitted among services.
There are two applications:
- Ping application that will invoke the Pong application using the RestClient and will return the expected "ping pong" output.
- Pong application that will return the "pong" output.

We're covering:
- Among REST services
- Among Reactive services
- With REST service to gRPC (this is not working... should be an issue? If not, I guess we should delete this).

Note that we don't need a running Jaeger to see this functionality working. This was odd to me at first and perhaps we should verify that indeed Jaeger is getting the metrics. Let me know your thoughts about this. @rsvoboda @mjurc

### DIFF
--- a/010-quarkus-opentracing-reactive-grpc/src/test/java/io/quarkus/qe/AbstractPingPongResourceTest.java
+++ b/010-quarkus-opentracing-reactive-grpc/src/test/java/io/quarkus/qe/AbstractPingPongResourceTest.java
@@ -1,0 +1,70 @@
+package io.quarkus.qe;
+
+import static io.restassured.RestAssured.given;
+import static org.awaitility.Awaitility.await;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.containsString;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.concurrent.TimeUnit;
+
+import org.apache.http.HttpStatus;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.junit.jupiter.api.Test;
+
+import io.quarkus.qe.containers.JaegerTestResource;
+import io.quarkus.test.common.QuarkusTestResource;
+
+@QuarkusTestResource(JaegerTestResource.class)
+public abstract class AbstractPingPongResourceTest {
+
+    private static final String PING_ENDPOINT = "/%s-ping";
+    private static final String PONG_ENDPOINT = "/%s-pong";
+
+    private static final String PING_RESOURCE = "PingResource";
+    private static final String PONG_RESOURCE = "PongResource";
+
+    @ConfigProperty(name = JaegerTestResource.JAEGER_API_ENDPOINT, defaultValue = "http://localhost:16686/api/traces")
+    String jaegerEndpoint;
+
+    @Test
+    public void testPingPong() {
+        // When calling ping, the rest will invoke also the pong rest endpoint.
+        given()
+                .when().get(pingEndpoint())
+                .then().statusCode(HttpStatus.SC_OK)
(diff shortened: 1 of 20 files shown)


010-quarkus-opentracing-reactive-grpc/src/test/java/io/quarkus/qe/GrpcPingPongResourceTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+import org.junit.jupiter.api.Disabled;
+
+import io.quarkus.qe.pong.GrpcPongService;
+import io.quarkus.test.junit.QuarkusTest;
+
+/**
+ * Disabled because OpenTracing does not work with REST -> gRPC.
+ */
+@Disabled
```
Sgitario, line 15: This test is not working: the trace headers are not transmitted to gRPC endpoint. I can see some bits that could be done to enable this integration (https://github.com/grpc-ecosystem/grpc-opentracing/tree/master/java).

However, taking into account that OpenTracing is deprecated. I think we should remove the gRPC test from here. What do you think? @rsvoboda @mjurc

rsvoboda: `@Disabled` may optionally be declared with a reason to document why the annotated test class or test method is disabled.

rsvoboda: I would keep the test in place, we can recheck it with OpenTelemetry

And please open issue for this in https://github.com/quarkusio/quarkus/issues so this gets discussed with developers

Sgitario: Raised https://github.com/quarkusio/quarkus/issues/13224 and updated the test to link this issue.

Sgitario:
@rsvoboda I managed to verify that the Jaeger service is indeed being updated just doing:

```java
// Then Jaeger is invoked
await().atMost(5, TimeUnit.SECONDS).untilAsserted(() -> given()
        .when().get("http://localhost:16686/api/traces/" + pingTraceId)
        .then().statusCode(HttpStatus.SC_OK)
        .and().body(allOf(containsString(PingResource.class.getName()), containsString(PongResource.class.getName()))));
```

Should this assertion be part of this test? What do you think?

rsvoboda:
You mean to pull in Jaeger testcontainer and check with the real backing service? If so, +1 for it.

Sgitario:
> You mean to pull in Jaeger testcontainer and check with the real backing service? If so, +1 for it.

Added.

010-quarkus-opentracing-reactive-grpc/src/test/java/io/quarkus/qe/AbstractPingPongResourceTest.java
```diff
@@ -0,0 +1,70 @@
...
+public abstract class AbstractPingPongResourceTest {
+
+    private static final String PING_ENDPOINT = "/%s-ping";
+    private static final String PONG_ENDPOINT = "/%s-pong";
+
+    private static final String PING_RESOURCE = "PingResource";
+    private static final String PONG_RESOURCE = "PongResource";
+
+    @ConfigProperty(name = JaegerTestResource.JAEGER_API_ENDPOINT, defaultValue = "http://localhost:16686/api/traces")
+    String jaegerEndpoint;
```
Sgitario, line 29: This property is injected once the Jaeger container is up and running, but I had to provide a default value.

rsvoboda (approved):

PR #46, opened 2020-11-19
Add test Infinispan server SSL/TLS secure connection

kshpak:
Test secure connection with Infinispan server.
Client and server use the same keystore/truststore file for authentication.
Native mode is working correctly. Secure communication was checked.
The first idea was to update the existing module `008-infinispan-getting-started` but through Testcontainers there was an issue with the TLS connection (server refused to send TLS Server Hello message). 
Moreover, the new module does not require client authentication on the server side through "username" and "password" parameters (i.e. the primary case in the `008-infinispan-getting-started` module), therefore I stick to the new module creation


012-infinispan-secure-server/pom.xml
```diff
@@ -0,0 +1,103 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
rsvoboda, line 1: Use `013-` prefix, Pablo uses 012 in his PR which arrived earlier

rsvoboda (changes requested):
> The first idea was to update the existing module 008-infinispan-getting-started but through Testcontainers there was an issue with the TLS connection (server refused to send TLS Server Hello message). 

This was good plan and I prefer this plan. Currently this PR mimics stuff in Quarkus TS - e.g. https://github.com/quarkusio/quarkus/blob/1.7.5.Final/integration-tests/infinispan-client/src/test/java/io/quarkus/it/infinispan/client/InfinispanServerTestResource.java#L36

Please investigate options for configuration of testcontainers based Infinispan instance. Don't be afraid to ask on Zulip (cc https://github.com/wburns and/or https://github.com/karesti) when you get stuck.

012-infinispan-secure-server/pom.xml
kshpak: ok, got you

rsvoboda:
Surpassed by https://github.com/quarkus-qe/beefy-scenarios/pull/49

PR #52, opened 2020-12-02
panache-rest-flyway: add many-to-one relationship with string id

Sgitario:
Added a new check to cover the many-to-one relationship in the Panache REST flyway module.


Sgitario:
@pjgg can you check what I did? For me, this scenario worked as expected.

pjgg:
Works as expected!.

pjgg (approved):

PR #54, opened 2020-12-03
Added module OIDC with RestClient

Sgitario:
Module that covers the RestClient and OIDC integration using Keycloak container.
There are two applications:
- Ping application that will invoke the Pong application using the RestClient and will return the expected "ping pong" output.
- Pong application that will return the "pong" output.

### DIFF
--- a/013-quarkus-oidc-restclient/README.md
+++ b/013-quarkus-oidc-restclient/README.md
@@ -1,0 +1,15 @@
+Module that covers the RestClient integration with OIDC extension using Keycloak container.
+There are two applications:
+- Ping application that will invoke the Pong application using the RestClient and will return the expected "ping pong" output.
+- Pong application that will return the "pong" output.
+
+Test cases:
+- When calling `/ping` or `/pong` endpoints without bearer token, then it should return 401 Unauthorized. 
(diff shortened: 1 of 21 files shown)


013-quarkus-oidc-restclient/README.md
```diff
@@ -0,0 +1,4 @@
+Module that covers the RestClient and OIDC integration using Keycloak container.
+There are two applications:
+- Ping application that will invoke the Pong application using the RestClient and will return the expected "ping pong" output.
+- Pong application that will return the "pong" output.
```
rsvoboda, line 4: Can you add case
A) when pong endpoint is accessed 
 - without expected Bearer token
 - with incorrect Bearer token

B) when ping endpoint is accessed without expected token (not sure if possible in this maven module or new module would be required)
?

Also please mention rest / reactive / lookup-auth variants of the tests

rsvoboda (changes requested):

013-quarkus-oidc-restclient/README.md
Sgitario: I've added a couple of more scenarios to cover the incorrect bearer cases and updated the readme. 
Let me know if it looks good now.

rsvoboda (approved):

PR #59, opened 2020-12-22
Added mysql and panache

Sgitario:
Added a module that test whether we can setup a REST API using ORM Panache and a multi-tenant data source configuration with Agroal extension and MySQL database.
These changes verify https://issues.redhat.com/browse/QUARKUS-549.

### DIFF
--- a/014-quarkus-panache-with-transactions-xa/README.md
+++ b/014-quarkus-panache-with-transactions-xa/README.md
@@ -1,0 +1,1 @@
+Module that test whether we can setup a REST API using ORM Panache and a datasource with XA transactions using a MySQL database.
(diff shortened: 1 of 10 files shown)


Sgitario:
I verified that this is not working using 1.7.5.Final upstream version and working fine using 1.7.6.Final for jvm and native both.

rsvoboda:
I miss highlight of the fact that XA is the important piece here.

Could you please rename the module to include `xa` ++ enhance README to include it too?

Sgitario:
> I miss highlight of the fact that XA is the important piece here.
> 
> Could you please rename the module to include `xa` ++ enhance README to include it too?

Done, let me know if it looks better to you know.

rsvoboda (approved):
Thanks for the update, waiting for CI to finish.

PR #64, opened 2020-12-23
Possible approach to test QUARKUS-608

Sgitario:


015-quarkus-container-configuration/pom.xml
```diff
@@ -0,0 +1,23 @@
...
+        <artifactId>beefy-scenarios</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>015-quarkus-container-configuration</artifactId>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-container-image-deployment</artifactId>
```
rsvoboda, line 19:
```suggestion
            <artifactId>quarkus-container-image</artifactId>
```

Deployment artifacts are not intended for end users

015-quarkus-container-configuration/src/test/java/io/quarkus/qe/ContainerImageTest.java
```diff
@@ -0,0 +1,114 @@
...
+        whenPublishImageInfo();
+        thenImageIs("group-with-space/" + APP_NAME + ":" + APP_VERSION);
+    }
+
+    private void givenNoUserName() {
+        givenProperty(USER_NAME_PROPERTY, StringUtils.EMPTY);
+    }
+
+    private void givenUserName(String value) {
+        givenProperty(USER_NAME_PROPERTY, value);
```
rsvoboda, line 82: Should be the property unset / set to the value before test execution?

rsvoboda (commented):
How does this module behave in native mode?

015-quarkus-container-configuration/pom.xml
Sgitario: I know, but I'm using internal classes like ContainerImageConfig and ContainerImageProcessor.

015-quarkus-container-configuration/src/test/java/io/quarkus/qe/ContainerImageTest.java
Sgitario: We should, yep

Sgitario:
@rsvoboda It's not working in native mode (it would need the skip flag in the maven configuration). However, this is not intended to be merged, just to demonstrate a possible approach to test different values/configuration in the same test suite. 
Feel free to not consider this PR if this is not the right approach!

015-quarkus-container-configuration/src/test/java/io/quarkus/qe/ContainerImageTest.java
```diff
@@ -0,0 +1,114 @@
...
+        givenUserName("user surname");
+        whenPublishImageInfo();
+        thenImageIs("user-surname/" + APP_NAME + ":" + APP_VERSION);
+    }
+
+    @Test
+    public void shouldNotFailWhenSpacesInGroupProperty() {
+        givenProperty(GROUP_PROPERTY, "group with space");
+        whenPublishImageInfo();
+        thenImageIs("group-with-space/" + APP_NAME + ":" + APP_VERSION);
```
jsmrcka, line 74: Given the nature of the fix, I think this test should always fail.

rsvoboda:
After few iterations we decided that upstream TS is the best place to put the coverage, inspired by this PR.

https://github.com/quarkusio/quarkus/blob/master/extensions/container-image/deployment/src/test/java/io/quarkus/container/image/deployment/ContainerImageConfigEffectiveGroupTest.java looks like a good fit to add new coverage

@jsmrcka please proceed this way

rsvoboda:
Jose Carvajal Hilario, 24 mins
The problem is that unit testing of ContainerImageConfig is not enough as when it's checking whether the final image repository is correct is here: ContainerImageProcessor
Concretely, in if (!ImageReference.isValidTag(effectiveTag)) {
and if (!ImageReference.isValidRepository(repository)) {

=> maybe the right place for putting test into is under https://github.com/quarkusio/quarkus/tree/master/integration-tests

rsvoboda:
There are some tests under https://github.com/quarkusio/quarkus/tree/master/integration-tests/container-image

Sgitario:
Closing as the approach will be to include this coverage into the quarkus repository directly.

PR #65, opened 2020-12-23
Quarkus / Vertx - jwt coverage

pjgg:
### Quarkus / Vertx JWT (AuthN/Z) exploratory testing

* AuthN
    * Validate JWT token: iss, ext, iat, aud
    * Validate JWT signature
* AuthZ (pending)
* Quarkus Vertx Reactive Routes
* Reactive redis

PR #69, opened 2021-01-07
Feature/vertx jwt redis

pjgg:
Backport related master PR: #65


rsvoboda:
I received `2021-01-07 13:56:22,393 ERROR [io.qua.mut.run.MutinyInfrastructure] (vert.x-eventloop-thread-12) Mutiny had to drop the following exception: java.lang.IllegalArgumentException: Unsupported Redis connection string scheme [localhost]` when running against 1.10.5.Final and 1.11.0.CR1, but luckily Redis is not supported on product side.
Adding `redis://` prefix solves that trouble.

rsvoboda:
https://github.com/quarkusio/quarkus/pulls?q=is%3Apr+ConfigProperties+is%3Aclosed there were several changes to ConfigProperty handling.

The fact that nested config didn't work well in 1.7 means that people probably avoided it the same way you did in the last commit. And the fact that this code works when running against 1.10.5.Final and 1.11.0.CR1 confirms that users shouldn't see any regression when upgrading.

302-quarkus-vertx-jwt/src/test/java/io/quarkus/qe/vertx/resources/RedisResource.java
```diff
@@ -17,7 +17,7 @@
         redisContainer = new GenericContainer(DockerImageName.parse("redis:5.0.3-alpine")).withExposedPorts(6379);
```
rsvoboda, line 17: quite old version of redis

rsvoboda (approved):

PR #82, opened 2021-02-08
Remove Amazon Lambda Common due to legacy package

Sgitario:
The build was failing with the next exception:

```
[error]: Build step io.quarkus.deployment.pkg.steps.JarResultBuildStep#buildRunnerJar threw an exception: java.lang.RuntimeException: Extensions with conflicting package types. One extension requires uber-jar another requires legacy format
```

The extension that was using the legacy package is amazon-lambda-common and hence removing it.

### DIFF
--- a/002-quarkus-all-extensions/pom.xml
+++ b/002-quarkus-all-extensions/pom.xml
@@ -393,10 +393,12 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-amazon-kms</artifactId>
         </dependency>
-        <dependency>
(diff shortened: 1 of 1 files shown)


Sgitario:
The legacy package is done on purpose, so I don't think this is an issue and we should report it.

pjgg (approved):

rsvoboda:
It this Quarkus 1.11 issue too or just Quarkus 1.12+

Sgitario:
> It this Quarkus 1.11 issue too or just Quarkus 1.12+

Only Quarkus 1.12+

PR #99, opened 2021-02-19
Add new module for vertx and routes

Sgitario:
This module covers some basic scenarios around reactive routes in general and also:
- Validation on request params, request body and responses.

### DIFF
--- a/304-quarkus-vertx-routes/src/main/java/io/quarkus/qe/validation/annotations/UppercaseService.java
+++ b/304-quarkus-vertx-routes/src/main/java/io/quarkus/qe/validation/annotations/UppercaseService.java
@@ -1,0 +1,12 @@
+package io.quarkus.qe.validation.annotations;
+
+import javax.enterprise.context.ApplicationScoped;
+
+@ApplicationScoped
+public class UppercaseService {
+    private static final String UPPERCASE_PATTERN = "[A-Z]+";
+
+    public boolean isUppercase(String value) {
+        return value != null && value.matches(UPPERCASE_PATTERN);
+    }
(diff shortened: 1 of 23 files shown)


Sgitario:
@pjgg and @mjurc , this PR can be safely be ported to 1.11 branch if you consider so, just let me know.

304-quarkus-vertx-routes/src/main/java/io/quarkus/qe/validation/annotations/UppercaseService.java
```diff
@@ -0,0 +1,12 @@
+package io.quarkus.qe.validation.annotations;
+
+import javax.enterprise.context.ApplicationScoped;
+
+@ApplicationScoped
+public class UppercaseService {
```
pjgg, line 6: This class is not really necessary. You could move on these validations to `UppercaseValidator`

pjgg (changes requested):

304-quarkus-vertx-routes/src/main/java/io/quarkus/qe/validation/annotations/UppercaseService.java
Sgitario: I know it's not necessary, but I wanted to verify that we can inject services in our custom validators.

rsvoboda: aka intentional, resolving and will merge

rsvoboda (approved):

PR #104, opened 2021-02-25
Fix consul test for windows and macos

Sgitario:
Using network=host is not compatible with Windows and MacOS.
Using the latest image of Consul allows running consul without network=host, just mapping the port.

### DIFF
--- a/019-quarkus-consul/src/test/java/io/quarkus/qe/containers/ConsulContainer.java
+++ b/019-quarkus-consul/src/test/java/io/quarkus/qe/containers/ConsulContainer.java
@@ -1,0 +1,17 @@
+package io.quarkus.qe.containers;
+
+import java.time.Duration;
+
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
+
+public class ConsulContainer extends GenericContainer<ConsulContainer> {
+    private static final int PORT = 8500;
+
+    public ConsulContainer() {
+        super("quay.io/bitnami/consul:1.9.3");
+        waitingFor(new LogMessageWaitStrategy().withRegEx(".*Synced node info.*\\s"));
+        withStartupTimeout(Duration.ofMillis(60000));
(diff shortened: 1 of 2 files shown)


019-quarkus-consul/src/test/java/io/quarkus/qe/containers/ConsulContainer.java
```diff
@@ -0,0 +1,17 @@
...
+import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
+
+public class ConsulContainer extends GenericContainer<ConsulContainer> {
+    private static final int PORT = 8500;
+
+    public ConsulContainer() {
+        super("quay.io/bitnami/consul:1.9.3");
+        waitingFor(new LogMessageWaitStrategy().withRegEx(".*Synced node info.*\\s"));
+        withStartupTimeout(Duration.ofMillis(60000));
+        addFixedExposedPort(PORT, PORT);
```
Sgitario, line 15: I had to create the ConsulContainer because I can't inject a random port into the running test (the `inject` method is not called when the test is not annotated with `@QuarkusTest`: I will see if this is something we can fix).

Sgitario:
@rsvoboda can you double check that the test is not working for you on MacOS?

rsvoboda (approved):
`mvn clean verify -pl 019-quarkus-consul/ -Dnative` passes now, thanks for the fix

rsvoboda:
`[INFO] 014-quarkus-panache-agroal ......................... FAILURE [ 53.005 s]` :/ not related to this PR, but probably worth investigation

I will merge the PR

Sgitario:
> `[INFO] 014-quarkus-panache-agroal ......................... FAILURE [ 53.005 s]` :/ not related to this PR, but probably worth investigation
> 
> I will merge the PR

This issue is caused by this commit: https://github.com/quarkusio/quarkus/commit/8c76e39060f6b288eaaeeb28bc477b33a972a40f 
I will investigate it further and report it.

PR #109, opened 2021-03-01
Update test class names to be unique for Polarion

kshpak:


rsvoboda (changes requested):
Adjust also IT tests - e.g. NativeReactivePingPongResourceIT are there twice

kshpak:
@rsvoboda updated. Better to pull changes to see the difference, the log is kinda messy

rsvoboda (approved):

PR #110, opened 2021-03-01
JPA mapping values coverage

pjgg:
The aim of this PR is to cover JPA `convert` annotation, motivated by this [issue](https://github.com/quarkusio/quarkus/issues/13234)

We have created a new field over the `Book` entity called `comments`. The type of this field is a `mutable type (a Map)`, and is converted to a JSON String through a JPA `convert` annotation.

### DIFF
--- a/601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/BookRepository.java
+++ b/601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/BookRepository.java
@@ -1,29 +1,31 @@
 package io.quarkus.qe.spring.data;
 
-import io.quarkus.qe.spring.data.model.Book;
 import java.util.List;
+
 import org.springframework.data.jpa.repository.Query;
-import org.springframework.data.repository.Repository;
+import org.springframework.data.repository.CrudRepository;
 import org.springframework.data.repository.query.Param;
 
-public interface BookRepository extends Repository<Book, Integer> {
+import io.quarkus.qe.spring.data.model.Book;
 
-    // issue 9192
+public interface BookRepository extends CrudRepository<Book, Integer> {
+
+    //This is for regression test for https://github.com/quarkusio/quarkus/pull/9192
     @Query(value = "SELECT b.publicationYear FROM Book b where b.bid = :bid")
     int customFindPublicationYearPrimitive(@Param("bid") Integer bid);
 
-    // issue 9192
+    //This is for regression test for https://github.com/quarkusio/quarkus/pull/9192
     @Query(value = "SELECT b.publicationYear FROM Book b where b.bid = :bid")
     Integer customFindPublicationYearObject(@Param("bid") Integer bid);
 
-    // issue 9192
+    //This is for regression test for https://github.com/quarkusio/quarkus/pull/9192
     @Query(value = "SELECT b.isbn FROM Book b where b.bid = :bid")
     long customFindPublicationIsbnPrimitive(@Param("bid") Integer bid);
 
-    // issue 9192
+    //This is for regression test for https://github.com/quarkusio/quarkus/pull/9192
     @Query(value = "SELECT b.isbn FROM Book b where b.bid = :bid")
     Long customFindPublicationIsbnObject(@Param("bid") Integer bid);
 
-    // issue QUARKUS-525
+    //This is for regression test for https://github.com/quarkusio/quarkus/pull/13015
     List<Book> findByPublisherAddressZipCode(@Param("zipCode") String zipCode);
 }
--- a/601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/BookResource.java
+++ b/601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/BookResource.java
@@ -1,11 +1,15 @@
 package io.quarkus.qe.spring.data;
 
-import io.quarkus.qe.spring.data.model.Book;
 import java.util.List;
+
 import javax.ws.rs.GET;
+import javax.ws.rs.NotFoundException;
+import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
+
+import io.quarkus.qe.spring.data.model.Book;
 
 @Path("/book")
 public class BookResource {
@@ -14,6 +18,15 @@ public class BookResource {
 
     public BookResource(BookRepository bookRepository) {
         this.bookRepository = bookRepository;
+    }
+
+    @PUT
+    @Produces("application/json")
+    @Path("/{id}")
+    public Book updateBook(@PathParam("id") Integer id, final Book book) {
+        verifyBookExists(id);
+        book.setBid(id);
+        return bookRepository.save(book);
     }
 
     @GET
@@ -47,8 +60,13 @@ public class BookResource {
     @GET
     @Path("/publisher/zipcode/{zipCode}")
     @Produces("application/json")
-    public List<Book> findBooksByPublicationYear(@PathParam("zipCode") String zipCode) {
+    public List<Book> findBooksByZipCode(@PathParam("zipCode") String zipCode) {
         return bookRepository.findByPublisherAddressZipCode(zipCode);
     }
 
+    private void verifyBookExists(Integer id) {
+        if (!bookRepository.existsById(id)) {
+            throw new NotFoundException(String.format("book with id=%d was not found", id));
+        }
+    }
 }
--- a/601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/model/MapStringConverter.java
+++ b/601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/model/MapStringConverter.java
@@ -1,0 +1,38 @@
+package io.quarkus.qe.spring.data.model;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Optional;
+
+import javax.persistence.AttributeConverter;
+
+import org.jboss.logging.Logger;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+class MapStringConverter implements AttributeConverter<Map<String, String>, String> {
+    private static final String EMPTY = "{}";
+    private final Logger logger = Logger.getLogger(MapStringConverter.class);
+    private final ObjectMapper objectMapper = new ObjectMapper();
+
+    @Override
+    public String convertToDatabaseColumn(Map<String, String> attribute) {
+        try {
+            return objectMapper.writeValueAsString(attribute);
+        } catch (IOException e) {
+            logger.error("Error while converting Map to JSON String: ", e);
+            return "";
+        }
+    }
+
+    @Override
+    public Map<String, String> convertToEntityAttribute(String dbData) {
+        try {
+            return objectMapper.readValue(Optional.ofNullable(dbData).orElse(EMPTY), Map.class);
+        } catch (Exception e) {
+            logger.error("Error while converting JSON String to Map: ", e);
+            return Collections.emptyMap();
+        }
+    }
+}
--- a/601-spring-data-primitive-types/src/test/java/io/quarkus/qe/spring/data/BookResourceTest.java
+++ b/601-spring-data-primitive-types/src/test/java/io/quarkus/qe/spring/data/BookResourceTest.java
@@ -1,18 +1,24 @@
 package io.quarkus.qe.spring.data;
 
+import static io.restassured.RestAssured.given;
+import static io.restassured.RestAssured.when;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.empty;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.core.Is.is;
+import static org.hamcrest.core.IsNot.not;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import org.junit.jupiter.api.Test;
+
 import io.quarkus.qe.spring.data.model.Book;
 import io.quarkus.test.junit.QuarkusTest;
 import io.restassured.http.ContentType;
 import io.restassured.response.Response;
-import java.util.Arrays;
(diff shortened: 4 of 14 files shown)


601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/BookRepository.java
```diff
@@ -3,10 +3,11 @@
 import io.quarkus.qe.spring.data.model.Book;
 import java.util.List;
 import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.CrudRepository;
 import org.springframework.data.repository.Repository;
 import org.springframework.data.repository.query.Param;
 
-public interface BookRepository extends Repository<Book, Integer> {
+public interface BookRepository extends Repository<Book, Integer>, CrudRepository<Book, Integer> {
```
jsmrcka, line 10: This seems redundant, `CrudRepository` already extends `Repository`.
```suggestion
public interface BookRepository extends CrudRepository<Book, Integer> {
```

601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/BookResource.java
```diff
@@ -51,4 +62,9 @@ public Long customFindPublicationIsbnObject(@PathParam("bid") Integer bid) {
         return bookRepository.findByPublisherAddressZipCode(zipCode);
     }
 
+    private void verifyFruitExists(Integer id) {
```
jsmrcka, line 65:
```suggestion
    private void verifyBookExists(Integer id) {
```
Although I must admit that some books can be just as fresh and delicious. 😊

601-spring-data-primitive-types/src/test/java/io/quarkus/qe/spring/data/BookResourceTest.java
```diff
@@ -48,11 +54,44 @@ void testCustomFindPublicationIsbnObjectLong() {
...
+        Map<String, String> comment = Collections.singletonMap("Shakespeare", "Lorem Ipsum Lorem Ipsum");
+        book.setComments(comment);
+        book = updateBook(book);
+
+        assertThat(book.getComments().size(), equalTo(1));
+        assertThat(book.getComments(), equalTo(comment));
+
+        // Update the previous comment
+        Map<String, String> commentUpdated = Collections.singletonMap("Shakespeare", "Lorem Ipsum Lorem Ipsum novus");
+        book.setComments(comment);
```
jsmrcka, line 77: If I understand it right, here should be the updated comment.
```suggestion
        book.setComments(commentUpdated);
```

601-spring-data-primitive-types/src/test/java/io/quarkus/qe/spring/data/BookResourceTest.java
```diff
@@ -48,11 +54,44 @@ void testCustomFindPublicationIsbnObjectLong() {
     // QUARKUS-525
     @Test
     void testFindBooksByPublisherZipCode() {
+        List<Book> books = retrieveBooksByZipcode();
+        books.stream().forEach(book -> assertThat(book.getPublisherAddress().getZipCode(), is("28080")));
+    }
+
+    // https://github.com/quarkusio/quarkus/issues/13234
+    @Test
+    @Disabled
```
jsmrcka, line 63: Why disabled?

jsmrcka (changes requested):

601-spring-data-primitive-types/src/test/java/io/quarkus/qe/spring/data/BookResourceTest.java
dreab8: Hi @jsmrcka, yes this is the cause of the test failure

601-spring-data-primitive-types/src/test/java/io/quarkus/qe/spring/data/BookResourceTest.java
pjgg: Was my fault!

jsmrcka (approved):

601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/BookRepository.java
```diff
@@ -1,12 +1,14 @@
...
-import org.springframework.data.repository.Repository;
+import org.springframework.data.repository.CrudRepository;
 import org.springframework.data.repository.query.Param;
 
-public interface BookRepository extends Repository<Book, Integer> {
+import io.quarkus.qe.spring.data.model.Book;
+
+public interface BookRepository extends CrudRepository<Book, Integer> {
 
     // issue 9192
```
kshpak, line 13: Seems like the issue is gone, please update that class. The same goes for QUARKUS-525, search and remove respective comments in classes

pjgg: Well I think, that keep the reference issue there in the code it's important in order to understand why is this code there. But is true that we should avoid internal references as `QUARKUS-525` and keep public references. So I've replaced, `QUARKUS-525` by `issue 13015`

kshpak:
BookRepositoryTest remove unused import org.hibernate.annotations.SQLUpdate;

601-spring-data-primitive-types/src/test/java/io/quarkus/qe/spring/data/BookResourceTest.java
```diff
@@ -48,11 +54,43 @@ void testCustomFindPublicationIsbnObjectLong() {
     // QUARKUS-525
     @Test
     void testFindBooksByPublisherZipCode() {
+        List<Book> books = retrieveBooksByZipcode();
+        books.stream().forEach(book -> assertThat(book.getPublisherAddress().getZipCode(), is("28080")));
+    }
+
+    // https://github.com/quarkusio/quarkus/issues/13234
```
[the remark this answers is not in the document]
kshpak: Delete unused import from this class "org.junit.jupiter.api.Disabled"

601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/BookRepository.java
kshpak: Change it to: // issue 9192; resolved; regression test

Sgitario: Please, add "TODO" and the fully qualified url of the issue, this way is much easier to search.

601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/BookRepository.java
```diff
@@ -24,6 +26,6 @@
     @Query(value = "SELECT b.isbn FROM Book b where b.bid = :bid")
     Long customFindPublicationIsbnObject(@Param("bid") Integer bid);
 
-    // issue QUARKUS-525
+    // issue 13015
```
Sgitario, line 29: Please, add "TODO" and the fully qualified url of the issue, this way is much easier to search.

601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/model/MapStringConverter.java
```diff
@@ -0,0 +1,38 @@
...
+
+import javax.persistence.AttributeConverter;
+
+import org.jboss.logging.Logger;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+class MapStringConverter implements AttributeConverter<Map<String, String>, String> {
+    private final Logger logger = Logger.getLogger(MapStringConverter.class);
+    private static final String EMPTY = "{}";
```
Sgitario, line 16: Static fields order should go before instance fields

601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/model/MapStringConverter.java
```diff
@@ -0,0 +1,38 @@
...
+        }
+    }
+
+    @Override
+    public Map<String, String> convertToEntityAttribute(String dbData) {
+        try {
+            return objectMapper.readValue(Optional.ofNullable(dbData).orElse(EMPTY), Map.class);
+        } catch (Exception e) {
+            logger.error("Error while converting JSON String to Map: ", e);
+            return new HashMap<>();
```
Sgitario, line 35: Use Collections.emptyMap.

601-spring-data-primitive-types/src/test/java/io/quarkus/qe/spring/data/BookResourceTest.java
```diff
@@ -48,11 +54,43 @@ void testCustomFindPublicationIsbnObjectLong() {
     // QUARKUS-525
```
Sgitario, line 54: Please, add "TODO" and the fully qualified url of the issue, this way is much easier to search.

601-spring-data-primitive-types/src/test/java/io/quarkus/qe/spring/data/BookResourceTest.java
```diff
@@ -5,11 +5,17 @@
 import io.restassured.http.ContentType;
 import io.restassured.response.Response;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
+
```
Sgitario, line 14: There are two empty lines here.

601-spring-data-primitive-types/src/test/java/io/quarkus/qe/spring/data/BookResourceTest.java
```diff
@@ -48,11 +54,43 @@ void testCustomFindPublicationIsbnObjectLong() {
...
+        book.setComments(commentUpdated);
+        book = updateBook(book);
+        assertThat(book.getComments().size(), equalTo(1));
+        assertThat(book.getComments(), equalTo(commentUpdated));
+    }
+
+    private Book updateBook(Book book) {
+        book = given().contentType(ContentType.JSON).body(book).when().put("/book/" + book.getBid()).then()
+                .statusCode(200).contentType(ContentType.JSON).extract().response().getBody().as(Book.class);
+        return book;
```
Sgitario, line 85: Return the book in a single line.

601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/BookRepository.java
kshpak: @Sgitario as I understand this issue is already solved and is used to see regression. I guess there is no need for TODO, do you agree? I agree that a fully qualified name is better

601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/BookResource.java
```diff
@@ -51,4 +62,9 @@ public Long customFindPublicationIsbnObject(@PathParam("bid") Integer bid) {
         return bookRepository.findByPublisherAddressZipCode(zipCode);
     }
 
+    private void verifyBookExists(Integer id) {
```
kshpak, line 65: Method above ^^^ i.e. public List<Book> findBooksByPublicationYear() actually searching by zip code, please change method name

601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/BookRepository.java
pjgg: This `issue` is just a reference in order to track, what this test is trying to cover. Is not a TODO. 
I will change `issue` by `githubRef`.

601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/BookRepository.java
Sgitario: If the issue is already solved, I would avoid adding any links at all.

601-spring-data-primitive-types/src/main/java/io/quarkus/qe/spring/data/BookRepository.java
Sgitario: I don't really like to have links or github references to things that are already resolved.
Everything we have in the tests are motivated by either an issue/feature, so we don't need to individually track any code snippets.
What do you think? cc @mjurc

pjgg: I think that our scenarios are sometimes very weird, and are easy to understand if you have a reference to the issue that is covering. Is like your context. Even more, if we use one scenario to cover different issues along the time.

pjgg: For example, could you tell me what is wrong, or what this test is trying to accomplish on this scenario:

` List<Book> findByPublisherAddressZipCode(@Param("zipCode") String zipCode);` 

This line is covering [a specific case](https://github.com/quarkusio/quarkus/pull/13015). 

If we drop the reference, somebody could change this line or some other stuff from the scenario, and maybe the test that is behind will also pass with the new "scenario" because the result is the same. But the aim of the test is totally broken.

mjurc: I don't have a strong opinion about their removal. Maybe some more obvious line to, well, make it more obvious it's a regression check for an issue?

Like, eg, `// This is a regression check for https://www.myissue`.

mjurc: I think this may be especially useful in tests which may otherwise not make much sense, actually.

mjurc (changes requested):
I would like the regression comments to be more clear.

Otherwise, I am fine with this PR.

mjurc (approved):
LGTM

Sgitario (approved):

PR #116, opened 2021-03-04
Disable 002 module in Native

Sgitario:
This PR is motivated because we can't exclude modules via Maven argument for the Quarkus ecosystem.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,6 @@
 
   <modules>
       <module>001-quarkus-getting-started-with-jaxrs</module>
-      <module>002-quarkus-all-extensions</module>
       <module>003-quarkus-many-extensions</module>
       <module>004-quarkus-HHH-and-HV</module>
       <module>005-quarkus-and-maven-profiles</module>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -144,6 +143,16 @@
     </plugins>
   </build>
   <profiles>
+    <profile>
+      <id>default</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
+      </activation>
+      <modules>
+        <!-- This module only works when running it in JVM mode -->
```
rsvoboda, line 152: Maybe mention that it needs 10+ GB of RAM for native-image process?

It works if you have enough memory ...

pom.xml
```diff
@@ -144,6 +143,16 @@
     </plugins>
   </build>
   <profiles>
+    <profile>
+      <id>default</id>
```
rsvoboda, line 147: use more descriptive name, something like `high-memory-needed`

pom.xml
Sgitario: At the moment, It's not working, do you want me to create another pull request with the changes needed to make the 002 scenario work in Native?

rsvoboda: pls create PR to make it work

Sgitario: https://github.com/quarkus-qe/beefy-scenarios/pull/118

pom.xml
Sgitario: Renamed

rsvoboda (approved):

PR #124, opened 2021-03-09
Add swagger-ui to non-appliaction endpoint test

pjgg:
Add swagger-ui to non-appliaction endpoint test


Sgitario (approved):

Sgitario:
This PR needs to be backported to 1.11 afterwards.

PR #138, opened 2021-03-15
Vertx-SQL: Reduce DB boots times

pjgg:
Move on 303-Vertx-SQL test cases to a common class in order to re-use DB instances and reduce boot times to 2:13 min.


Sgitario (commented):
This solution is using inheritance by adding implementations on interfaces (using default methods). I consider this a bad practice and I would suggest using composition instead. However, maybe this is a matter of dev personal style and it should not be a stopper for this PR, so I will delegate it to @mjurc if you don't mind.

Note that I tried to build this module locally and It's failing for me with:

```
[ERROR] Failures: 
[ERROR]   PostgresPoolTest.checkDbPoolTurnover:93 Missing 24325 events. ==> expected: <675> but was: <0>
```

But as It's not failing for Github CI, this must be something on my local. Please, check this too.

pjgg:
> This solution is using inheritance by adding implementations on interfaces (using default methods). I consider this a bad practice and I would suggest using composition instead. However, maybe this is a matter of dev personal style and it should not be a stopper for this PR, so I will delegate it to @mjurc if you don't mind.
> 
> Note that I tried to build this module locally and It's failing for me with:
> 
> ```
> [ERROR] Failures: 
> [ERROR]   PostgresPoolTest.checkDbPoolTurnover:93 Missing 24325 events. ==> expected: <675> but was: <0>
> ```
> 
> But as It's not failing for Github CI, this must be something on my local. Please, check this too.

Well, we could move on all the tests to several classes without state and without constructor but... this is an interface (from java 8). What do you think @mjurc

mjurc (approved):
LGTM. Thanks for the update.

PR #143, opened 2021-03-18
GH claim disk space

pjgg:
Claim some extra disk space to GH in order to avoid errors like :
```
❌ Docker environment should have more than 2GB free disk space
```

### DIFF
--- a/.github/ci-prerequisites.sh
+++ b/.github/ci-prerequisites.sh
@@ -1,0 +1,10 @@
+# Reclaim disk space, otherwise we only have 13 GB free at the start of a job
+echo "Reclaim disk space."
+df -h
+time docker rmi node:10 node:12 mcr.microsoft.com/azure-pipelines/node8-typescript:latest
+# That is 18 GB
+time sudo rm -rf /usr/share/dotnet
+# That is 1.2 GB
+time sudo rm -rf /usr/share/swift
+echo "Reclaim disk space end."
+df -h
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -17,6 +17,8 @@ jobs:
           key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
             ${{ runner.os }}-maven-
+      - name: Reclaim Disk Space
+        run: .github/ci-prerequisites.sh
       - name: Install JDK {{ matrix.java }}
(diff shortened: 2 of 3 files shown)


rsvoboda:
@Sgitario do we need the same change also for https://github.com/quarkus-qe/beefy-scenarios/blob/main/.github/workflows/quarkus-snapshot.yaml ? Or it triggers just the JVM run?

rsvoboda (approved):

rsvoboda:
Jose will prepare the PR for ecosystem ci directly.

.github/ci-prerequisites.sh
```diff
@@ -0,0 +1,7 @@
+# Reclaim disk space, otherwise we only have 13 GB free at the start of a job
+
```
rsvoboda, line 2: add `df -h` to see stats about disk usage before cleanup

.github/ci-prerequisites.sh
```diff
@@ -0,0 +1,7 @@
...
+time docker rmi node:10 node:12 mcr.microsoft.com/azure-pipelines/node8-typescript:latest
+# That is 18 GB
+time sudo rm -rf /usr/share/dotnet
+# That is 1.2 GB
+time sudo rm -rf /usr/share/swift
```
rsvoboda, line 7: add `df -h` to see stats about disk usage after the cleanup

rsvoboda (changes requested):

.github/ci-prerequisites.sh
pjgg: I've also add an echo message

.github/ci-prerequisites.sh
pjgg: Before: `/dev/root        84G   68G   16G  82% /`
After: `/dev/root        84G   43G   41G  51% /`

.github/ci-prerequisites.sh
```diff
@@ -0,0 +1,11 @@
+# Reclaim disk space, otherwise we only have 13 GB free at the start of a job
+
+time docker rmi node:10 node:12 mcr.microsoft.com/azure-pipelines/node8-typescript:latest
+echo "Reclaim disk space."
+df -h
```
rsvoboda, line 5: `df-h` should be before docker rmi execution

rsvoboda (approved):
Looks good, lets wait for CI to finish

PR #146, opened 2021-03-19
Reclaim disk space for native daily build too

rsvoboda:
Reclaim disk space for native daily build too


rsvoboda:
@pjgg fyi, we missed this cas

PR #150, opened 2021-03-22
Migrate 007 Scenario to 020-quarkus-http-non-application-endpoints

pjgg:
Goal: save native package time.
Desc: Openshift health probe scenario, could run with any other scenario and save one native package time in our monorepo.


rsvoboda (changes requested):
What's the motivation to use 004 as the target module?

I would put 007-openshift-yaml-health-probes based checks into 020-quarkus-http-non-application-endpoints

007-openshift-yaml-health-probes/README.md
```diff
@@ -1,7 +0,0 @@
-# Quarkus - OpenShift extension
```
rsvoboda, line 1: Would keep this file and point towards the new module where the coverage was migrated

rsvoboda (approved):

PR #151, opened 2021-03-22
Refactor kafka modules: Move 012 scenario to 301 scenario

pjgg:
Goal: save native package time.
We have moved Quarkus Kafka streams scenario to Quarkus Kafka-vertx.


rsvoboda:
We have daily runs on BeefyTS with Mandrel and GraalVM against Quarkus Master, we do not have the same in OpenShiftTS.
BeefyTS is also present in https://github.com/quarkusio/quarkus-ecosystem-ci.

Before merging this removal we need to make sure that the target repo has the same presence as BeefyTS.

rsvoboda:
@pjgg please look at `009-quarkus-infinispan-grpc-kafka` failure

pjgg:
> We have daily runs on BeefyTS with Mandrel and GraalVM against Quarkus Master, we do not have the same in OpenShiftTS.
> BeefyTS is also present in https://github.com/quarkusio/quarkus-ecosystem-ci.
> 
> Before merging this removal we need to make sure that the target repo has the same presence as BeefyTS.

Yes, I think that is a good idea that both modules remain on this project. I have joined both modules in a single one.

rsvoboda:
One concern I have is if we will be able to easily say if the problem we have is with Kafka Streams or with vanilla Kafka.

pjgg:
> One concern I have is if we will be able to easily say if the problem we have is with Kafka Streams or with vanilla Kafka.

Yes, we have different tests / tocpis. And Kafka itself is agnostic about ...if a topic is consumed from a stream of from a KafkaConnector/consumer.

rsvoboda (approved):

PR #155, opened 2021-03-23
Avoid mutiny/blocking grpc be under the same service in order to avoid service.name collisions.

pjgg:
Refactor .proto files in order to avoid service names colissions:

Error example:
```
java.lang.IllegalStateException: Duplicated gRPC service: helloworld.Greeter
```


Sgitario (approved):

Sgitario:
Another note: This issue happens when running the module in DEV mode: `mvn clean quarkus:dev` and it's not a regression issue (happens the same in 1.10.Final, for example). For some reason, running the tests does not spot it.

PR #156, opened 2021-03-30
Add test to cover Quarkus configuration properties

kshpak:
This test covers most of the config properties for Quarkus described in Configuring you Quarkus application guide
https://access.redhat.com/documentation/en-us/red_hat_build_of_quarkus/1.7/html-single/configuring_your_quarkus_applications/index

resources-config.json file should be there according to https://quarkus.io/guides/writing-native-applications-tips#supporting-native-in-your-application (without it, the errors are thrown on the following line
`InputStream in = CustomConfigSource.class.getResourceAsStream(PROPERTIES_FILE);`
)

Module has reach README file

### DIFF
--- a/022-quarkus-properties-config-all/pom.xml
+++ b/022-quarkus-properties-config-all/pom.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0"?>
+<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.quarkus.qe</groupId>
+        <artifactId>beefy-scenarios</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>022-quarkus-properties-config-all</artifactId>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-config-yaml</artifactId>
+        </dependency>
+    </dependencies>
+
+    <profiles>
+        <profile>
+            <id>native</id>
+            <properties>
+                <quarkus.package.type>native</quarkus.package.type>
+                <quarkus.native.additional-build-args>-H:ResourceConfigurationFiles=resources-config.json</quarkus.native.additional-build-args>
+            </properties>
+        </profile>
+    </profiles>
+</project>
--- a/022-quarkus-properties-config-all/.env
+++ b/022-quarkus-properties-config-all/.env
@@ -1,0 +1,1 @@
+WELCOME_MESSAGE=Welcome message from .env file
--- a/022-quarkus-properties-config-all/README.md
+++ b/022-quarkus-properties-config-all/README.md
@@ -1,0 +1,28 @@
+# Quarkus - Cover application properties setup
+This module tests different ways of configuring properties for Quarkus application
+
+1. Configuring custom user properties through nested static class (done for Antagonist)  
+1.1 Configuration properties located in the *application.properties* file
+
+2. Configuring custom user properties through set of interfaces (done for Protagonist)  
+2.1 Configuration properties located in the *application.yaml* file (this file takes precedence over the
+*application.properties* file)
+
+3. Configuring custom config source provider  
+3.1 File org.eclipse.microprofile.config.spi.ConfigSourceProvider must be created under **META-INF/services/**
+    and contains fully qualified names of implementations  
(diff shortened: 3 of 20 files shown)


022-quarkus-properties-config-all/src/main/java/org/acme/GreetingResource.java
```diff
@@ -0,0 +1,78 @@
...
+import org.acme.converter.KrustyEmail;
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+import javax.inject.Inject;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@Path("/hello")
+public class GreetingResource{
```
rsvoboda, line 19:
```suggestion
public class GreetingResource {
```
the same is on few other places

022-quarkus-properties-config-all/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+      <artifactId>quarkus-config-yaml</artifactId>
+    </dependency>
+  </dependencies>
+
+  <profiles>
+    <profile>
+      <id>native</id>
+      <properties>
+        <quarkus.package.type>native</quarkus.package.type>
+        <quarkus.native.additional-build-args>-H:ResourceConfigurationFiles=resources-config.json</quarkus.native.additional-build-args>
```
rsvoboda, line 29: No hint is needed for JVM mode, just for native mode ?

rsvoboda (changes requested):
More general comments:
 - use quarkus qe package name 
 - can be this test coverage merged with existing module - e.g. 001-quarkus-getting-started-with-jaxrs?  Motivation is to save time on native image compilation.
    - 001-quarkus-getting-started-with-jaxrs => 001-quarkus-jaxrs-and-config-properties

022-quarkus-properties-config-all/src/main/java/org/acme/GreetingResource.java
```diff
@@ -0,0 +1,78 @@
...
+    @Inject
+    Config config;
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    public String welcomeMessage(){
+        return config.getValue("welcome.message", String.class);
+    } // declared in .env
+
+
```
Sgitario, line 30: Double empty line here.

022-quarkus-properties-config-all/src/main/java/org/acme/GreetingResource.java
```diff
@@ -0,0 +1,78 @@
...
+    @GET
+    @Path("/antagonist/wife")
+    @Produces(MediaType.TEXT_PLAIN)
+    public String helloAntagonistWife() {
+        return antagonist.wife.name + " says: " + antagonist.wife.message;
+    }
+
+
+    @ConfigProperty(name = "restaurant.employees")
+    List<KrustyEmail> employeesEmails;
```
Sgitario, line 68: Fields should be defined before methods. Same with the other fields.

022-quarkus-properties-config-all/pom.xml
kshpak: right, just for native mode. without it tests still pass, but there are a bunch of errors printed before, which slowdown native execution. errors related to loading additional config source

kshpak:
All comments reflected in changes. I would prefer to leave it as a separate module

022-quarkus-properties-config-all/src/test/java/quarkus/qe/GreetingResourceTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+import io.quarkus.test.junit.QuarkusTest;
+import org.junit.jupiter.api.Test;
+
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.CoreMatchers.is;
+
+@QuarkusTest
+public class GreetingResourceTest {
```
rsvoboda, line 11: I would use something like `VariousConfigurationSourcesTest`

022-quarkus-properties-config-all/src/test/java/quarkus/qe/GreetingResourceTest.java
```diff
@@ -0,0 +1,69 @@
...
+    public void testEnvFile() {
+        given()
+          .when().get("/hello")
+          .then()
+             .statusCode(200)
+             .body(is("Welcome message from .env file"));
+    }
+
+    @Test
+    public void testProtagonistEndpoint() {
```
rsvoboda, line 23: Indicate what is the source of the configuration (application properties, yaml, resources-config.json etc) in the test name
Do not use the endpoint name, that doesn't hint the problematic area if the test fails

rsvoboda (changes requested):

022-quarkus-properties-config-all/src/test/java/quarkus/qe/GreetingResourceTest.java
kshpak: Good point. Changed

Sgitario (approved):

rsvoboda (approved):

rsvoboda:
waiting for ci ...

PR #158, opened 2021-03-31
Kafka AVRO native test

pjgg:
1. Push back Kafka Streams scenarios to module 012, because makes easy to test/debug
2. Upgrade apicurio-registry-utils-serde 1.3.2.Final (Required for Kafka AVRO)
3. Add Apicurio Rest client that support AVRO on native mode `quarkiverse-apicurio-registry-client`
4. Force `camel-quarkus-support-retrofit` to version `1.7.0`
5. Add StrimziKafkaIT (native test)
6. Some other changes as emit an event every 100ms in order to make the test scenario more resilience.


pom.xml
```diff
@@ -48,8 +49,10 @@
     <testcontainers.version>1.15.2</testcontainers.version>
     <strimzi.testcontainers.version>0.20.0</strimzi.testcontainers.version>
     <wiremock.version>2.27.2</wiremock.version>
-    <apicurio-registry-utils-serde.version>1.2.2.Final</apicurio-registry-utils-serde.version>
+    <apicurio-registry-utils-serde.version>1.3.2.Final</apicurio-registry-utils-serde.version>
     <confluent.kafka-avro-serializer.version>6.0.0</confluent.kafka-avro-serializer.version>
+    <version.quarkiverse.apicurio.registry.client>0.0.1</version.quarkiverse.apicurio.registry.client>
+    <camel-quarkus-support-retrofit.version>1.7.0</camel-quarkus-support-retrofit.version>
```
rsvoboda, line 55: The right version should really come from Quarkus deps

pom.xml
```diff
@@ -48,8 +49,10 @@
     <testcontainers.version>1.15.2</testcontainers.version>
     <strimzi.testcontainers.version>0.20.0</strimzi.testcontainers.version>
     <wiremock.version>2.27.2</wiremock.version>
-    <apicurio-registry-utils-serde.version>1.2.2.Final</apicurio-registry-utils-serde.version>
+    <apicurio-registry-utils-serde.version>1.3.2.Final</apicurio-registry-utils-serde.version>
```
rsvoboda, line 52: Could be this controlled by quarkiverse-apicurio-registry-client dependency or this is independent on client?

Asking because end user experience is quite poor, users need to figure out many things on their own instead of just one dependency that just works.

rsvoboda (approved):
Those two comments are about usability of AVRO bits, not about the test coverage.

There should be some followup action or discussions to address these concerns.

PR #160, opened 2021-04-07
Reproducer for quarkus/issues/15025: Quarkus keeps dead database connect

Sgitario:
Reproducer for https://github.com/quarkusio/quarkus/issues/15025: Quarkus keeps dead database connections in its connection pool.

The idea is to keep restarting the database and checking the idle connection pool. Before each restart, we need to call the health check endpoint to trigger the issue. 

This is working fine in 999-SNAPSHOT, but not in 1.13.0.Final (as expected as the fix is not part of this release):

Works:
```
mvn clean install
```

Not work:
```
mvn clean install -Dquarkus.platform.version=1.13.0.Final -Dquarkus-plugin.version=1.13.0.Final
```


rsvoboda:
Kyrylo is sick

Sgitario:
Adding "DO NOT MERGE" as it's failing in CI... I'm investigating this.

Sgitario:
Fixed by adding DB_CLOSE into the jdbc url. It seems that when we stop/start the database connection, this property is necessary.

PR #161, opened 2021-04-07
Kstream commit interval config

pjgg:
Currently, we are not setting any `kafka-streams.commit.interval.ms` so looks like by default this property is set to `30000 milliseconds`. 

The test is going to take less time to succeed if we reduce this commit interval.


rsvoboda:
Reducing commit interval ... is this just test case optimization or do you think there should be change of the default?

pjgg:
> Reducing commit interval ... is this just test case optimization or do you think there should be change of the default?

Test-case optimization

Sgitario (approved):

PR #163, opened 2021-04-08
Update Beefy TS to Quarkus 2.x (Vertx 4 - 999-SNAPSHOT)

Sgitario:
Changes:
- Change HttpMethod to use the one from vertx route:
The reactive routes signature have changed and need to use another HttpMethod enum. More info in here: https://github.com/quarkusio/quarkus/commit/72da8546b564aa4871ea16aa4bf5b930f0f9286b

- Regression issue: Infinispan and Kafka extensions stopped working at the same time, reported by https://github.com/quarkusio/quarkus/issues/16386

- The annotation `@Health` got removed. I changed it to `@Readiness`. I've also added a new test case for covering the `HealthGroup` annotation and reported https://github.com/quarkusio/quarkus/issues/16389

- Regression Test: Validation errors on Reactive routes are no longer propagated. Reported by https://github.com/quarkusio/quarkus/issues/16430

- Regression issue: SmallRye OpenAPI generates incorrect content type when using Spring Web annotations. Reported by https://github.com/quarkusio/quarkus/issues/16395

PR #164, opened 2021-04-09
Vertx 4 issue Reproducer

pjgg:
When I launch my app and make a query to `/issue` works as expected but
when I make the same query through restAssurance I got an exception.


Sgitario (changes requested):
Nack. This will be done as part of https://github.com/quarkus-qe/beefy-scenarios/pull/163.

PR #168, opened 2021-04-14
Fix module 302 in Native

Sgitario:
This module was failing in Native because this exception:

```
{"timestamp":1618384522931,"status":500,"error":"Internal Server Error","path":"/bladeRunner/","message":"Failed to encode as JSON: No serializer found for class io.vertx.ext.auth.impl.UserImpl and no properties discovered to create BeanSerializer (to avoid exception, disable SerializationFeature.FAIL_ON_EMPTY_BEANS)"}
```

Note that I tried to approach this error following this guide: https://quarkus.io/guides/writing-native-applications-tips#using-the-registerforreflection-annotation

This means that I tried:
- Adding `@RegisterForReflection(targets=UserImpl.class)`
- And adding the `reflection-config.json` file

Any of these two approaches worked, but disabling the failures on empty beans.


Sgitario:
I didn't report any issue as I don't think it's a bug in Quarkus, but because we're using the vertx-jwt.

rsvoboda:
CI is green, I want to understand if this is Quarkus bug or not

PR #174, opened 2021-04-15
Vertx Auth User serialization error on native mode (patch)

pjgg:
- Remove unused lib on Quarkus 2.x
- comment a conflict log line that is throwing a serialization error.


rsvoboda:
there is already https://github.com/quarkus-qe/beefy-scenarios/pull/168 to address native fail

I would keep `LOG.info("Ready for AuthZ. User: " + Json.encode(user));` in the code, I think people can have it in their codebase.

PR #182, opened 2021-04-22
Configure Github CI to run tests on Windows

Sgitario:
Configured only for:
- PRs (similarly as done for Linux)
- Daily (only graalvm)
- Daily Native: Only the module 003 as build native requires lots of disk space and it was failing with `Error: There is not enough space on the disk.`

Not done for Ecosystem.

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -1,10 +1,11 @@
 name: "Daily Build"
 on:
+  workflow_dispatch:
   schedule:
     - cron: '0 23 * * *'
 jobs:
-  build-jvm:
-    name: Daily build - JVM
+  quarkus-main-build:
+    name: Quarkus Main build
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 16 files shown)


rsvoboda:
`Could not find a valid Docker environment.` on Windows run.

Seems only Windows containers are supported - https://github.com/actions/virtual-environments/issues/1143#issuecomment-651178811

Sgitario:
We can verify the PR build as part of this PR and the Daily build in here: https://github.com/Sgitario/beefy-scenarios/actions/runs/809481747

Sgitario:
Failures will be addressed in:
- gRPC failures: https://github.com/quarkus-qe/beefy-scenarios/pull/197
- Quartz: https://github.com/quarkus-qe/beefy-scenarios/pull/198

Sgitario:
@rsvoboda finally, this is ready to be reviewed. You can take a look into a Daily run here: https://github.com/Sgitario/beefy-scenarios/actions/runs/816237455

rsvoboda (approved):

rsvoboda:
Let's merge it, skip of the tests could be handled via property, but that's not a bigiie.
In the future we may need to adjust the behavior based on provided flag, not just os detection.
The problem is on Windows which are running in VM, not general Windows problem.
That can be handled in the upcoming work.

PR #186, opened 2021-04-26
Apply format using Maven plugins

Sgitario:
The format will be automatically applied using the Maven plugins:
- formatter-maven-plugin - for sources
- impsort-maven-plugin - for imports
- xml-format-maven-plugin - for XML files (POM files)


Sgitario:
This PR uses the same approach as Quarkus upstream plus adding the format also for XML files. 
We've reverted the changes in https://github.com/quarkus-qe/beefy-scenarios/pull/180/files that were using Checkstyle to achieve the same purposes, but it was much restricted than now.

rsvoboda:
Can we have two separate commits pls? One for introduction of new plugins + profiles + GH Actions. One for the changes introduced by execution of these plugins. So we have some separation.

Sgitario:
> Can we have two separate commits pls? One for introduction of new plugins + profiles + GH Actions. One for the changes introduced by execution of these plugins. So we have some separation.

Done

rsvoboda (approved):

mjurc (approved):
Thanks for the PR, looks good to me!

Merging.

PR #187, opened 2021-04-26
Kafka AVRO class loader issue

pjgg:


kshpak (commented):
The test io.quarkus.qe.kafka.StrimziKafkaIT is failing in native

[Update] This patch fixes JVM mode, the test is still flaky in native mode. However, the native flakiness shouldn't be affected by this patch.

kshpak (approved):

PR #192, opened 2021-04-30
[1.11] Add debug native symbols generation test

pjgg:
depends on: #191


rsvoboda:
3 commits, please cherry pick just the one commit.

Test is passing even-though it should not.

To properly check the original issue we would need to make sure that the target machine doesn't have `objcopy` available.
We will ensure it on Windows platform. Another option is to check log file and search for the used command.

pjgg:
> We will ensure it on Windows platform.

I agree!

rsvoboda (approved):

PR #196, opened 2021-05-04
Update formatter rules to validate `io` packages, avoid import stars and fixed when selecting Maven profiles

Sgitario:
- Update import rules to validate io packages at the end
- Fix formatter when using Maven profiles
At the moment, it was failing when selecting Maven profiles because the `format` profile was unselected and hence the properties were not present.

- Add Checkstyle to avoid import statements that use the * notation.


Sgitario:
Failures will be fixed as part of https://github.com/quarkus-qe/beefy-scenarios/pull/197

Sgitario:
The dependant PRs have been merged and now the CI is green. This PR can now be reviewed. cc @rsvoboda

Sgitario:
@rsvoboda this is ready to be merged. I think it's better to not squash the commits.

rsvoboda (approved):

PR #197, opened 2021-05-04
Fix gRPC tests

Sgitario:
As part of https://github.com/quarkusio/quarkus/pull/16830, Quarkus is validating that gRPC services are annotated with `@GrpcService`.


Sgitario:
Note that there are some expected failures related to a deprecated Quartz property and that will be addressed in https://github.com/quarkus-qe/beefy-scenarios/pull/198

kshpak (approved):

PR #203, opened 2021-05-11
Agroal DB pool test coverage

pjgg:
- Add some test coverage in order to double-check Agroal DB pool management.

### DIFF
--- a/014-quarkus-panache-with-transactions-xa/README.md
+++ b/014-quarkus-panache-with-transactions-xa/README.md
@@ -16,3 +16,5 @@ Additional UserEntity is a simple JPA entity that was created with aim to avoid 
 and instead test the additional combination of JPA entity + PanacheRepository + PanacheRepositoryResource, where
 PanacheRepository is a facade class. Facade class can override certain methods to change the default behaviour of the
 PanacheRepositoryResource methods.
+
+- AgroalPoolTest, will cover how the db pool is managed in terms of IDLE-timeout, max connections and concurrency. 
--- a/014-quarkus-panache-with-transactions-xa/src/main/resources/application.properties
+++ b/014-quarkus-panache-with-transactions-xa/src/main/resources/application.properties
@@ -17,3 +17,16 @@ quarkus.datasource.with-xa.jdbc.transactions=xa
 
 quarkus.hibernate-orm.database.generation=drop-and-create
 quarkus.hibernate-orm.sql-load-script=import.sql
+
+%mysql_pool_test.quarkus.datasource.db-kind=mysql
+%mysql_pool_test.quarkus.datasource.username=sarah
+%mysql_pool_test.quarkus.datasource.password=connor
+%mysql_pool_test.quarkus.datasource.jdbc.url=jdbc:mysql://localhost:3306/db?currentSchema=test
+%mysql_pool_test.quarkus.datasource.jdbc.transactions=enabled
+%mysql_pool_test.quarkus.datasource.jdbc.max-size=3
+%mysql_pool_test.quarkus.datasource.jdbc.min-size=1
+%mysql_pool_test.quarkus.datasource.jdbc.background-validation-interval=1S
+%mysql_pool_test.quarkus.datasource.jdbc.idle-removal-interval=1S
+%mysql_pool_test.quarkus.datasource.jdbc.acquisition-timeout=60S
+%mysql_pool_test.quarkus.datasource.jdbc.validation-query-sql=SELECT CURRENT_TIMESTAMP
+%mysql_pool_test.quarkus.datasource.jdbc.new-connection-sql=SELECT CURRENT_TIMESTAMP
--- a/014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/containers/MysqlTestProfile.java
+++ b/014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/containers/MysqlTestProfile.java
@@ -1,0 +1,21 @@
+package io.quarkus.qe.containers;
+
+import java.util.Collections;
+import java.util.List;
+
+import io.quarkus.test.junit.QuarkusTestProfile;
+
+public class MysqlTestProfile implements QuarkusTestProfile {
+
+    public static final String PROFILE = "mysql_pool_test";
+
+    @Override
+    public String getConfigProfile() {
+        return PROFILE;
+    }
+
+    @Override
+    public List<TestResourceEntry> testResources() {
+        return Collections.singletonList(new TestResourceEntry(MySqlDatabaseTestResource.class));
+    }
+}
--- a/014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
+++ b/014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
@@ -1,0 +1,176 @@
+package io.quarkus.qe.dbpool;
+
+import static io.restassured.RestAssured.given;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.sql.Connection;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.time.Duration;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import javax.inject.Inject;
+import javax.persistence.EntityManager;
+import javax.persistence.Query;
+
+import org.apache.http.HttpStatus;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.RepeatedTest;
+import org.junit.jupiter.api.Test;
+
+import io.agroal.api.AgroalDataSource;
+import io.quarkus.qe.containers.MysqlTestProfile;
+import io.quarkus.qe.rest.data.UserRepository;
+import io.quarkus.test.junit.QuarkusTest;
+import io.quarkus.test.junit.TestProfile;
+import io.smallrye.mutiny.Multi;
+import io.smallrye.mutiny.helpers.test.AssertSubscriber;
+import io.smallrye.mutiny.tuples.Tuple2;
+
+/**
+ * The aim of these tests is verified agroal and entityManager pool management
+ * Some of these tests required some extra load, in order to reproduce concurrency issues.
+ */
+@QuarkusTest
+@TestProfile(MysqlTestProfile.class)
+public class AgroalPoolTest {
+
+    private final int CONCURRENCY_LEVEL = 20;
+
+    @Inject
+    EntityManager em;
+
+    @Inject
+    UserRepository users;
+
+    @Inject
+    AgroalDataSource agroalDataSource;
+
+    @ConfigProperty(name = "quarkus.datasource.jdbc.idle-removal-interval")
+    String idleSec;
+
+    @ConfigProperty(name = "quarkus.datasource.jdbc.max-size")
+    int datasourceMaxSize;
+
+    @ConfigProperty(name = "quarkus.datasource.jdbc.min-size")
+    int datasourceMinSize;
+
+    @Test
+    @Disabled("Pending Zulip conversation about quarkus.datasource.jdbc.idle-removal-interval")
+    public void idleTimeoutTest() throws InterruptedException {
+        makeApplicationQuery();
+        Thread.sleep(Duration.ofMillis(getIdleMs() + 1).toMillis());
+        assertEquals(1, activeConnections(), "agroalCheckIdleTimeout: Expected " + datasourceMinSize + " active connections");
+    }
+
+    @Test
+    public void poolTurnoverTest() throws InterruptedException {
+        final int events = 500;
+        final AtomicInteger max = new AtomicInteger(0);
+
+        AssertSubscriber<Tuple2<Long, Long>> subscriber = Multi.createBy().combining()
+                .streams(makeApplicationQueryAsync(events), activeConnectionsAsync(events))
+                .asTuple().subscribe().withSubscriber(AssertSubscriber.create(events));
+
+        subscriber.awaitItems(events).getItems().forEach(t -> {
+            Long activeCon = t.getItem2();
+            if (max.intValue() < activeCon) {
+                max.set(activeCon.intValue());
+            }
+            assertTrue(datasourceMaxSize >= activeCon, "More mysql SQL connections than pool max-size");
+            assertTrue(datasourceMinSize <= activeCon, "Less mysql SQL connections than pool min-size");
+        });
+    }
+
+    @RepeatedTest(500)
+    public void borderConditionBetweenIdleAndGetConnectionTest() {
+        AssertSubscriber<Integer> subscriber = Multi.createFrom().range(0, CONCURRENCY_LEVEL).flatMap(n -> Multi
+                .createFrom().ticks()
+                .every(Duration.ofMillis(getIdleMs() + 3))
+                .onOverflow().drop()
+                .onFailure().invoke(e -> Assertions.fail("Unexpected exception " + e.getMessage()))
+                .onItem().transform(i -> makeApplicationQuery()))
+                .subscribe()
+                .withSubscriber(AssertSubscriber.create(CONCURRENCY_LEVEL));
+
+        subscriber
+                .awaitItems(CONCURRENCY_LEVEL)
+                .getItems()
+                .forEach(statusCode -> assertEquals(statusCode, HttpStatus.SC_OK, "Unexpected Application response"));
+    }
+
+    @RepeatedTest(100)
+    public void concurrentLoadTest() {
+        Multi.createFrom()
+                .range(0, CONCURRENCY_LEVEL).subscribe()
+                .with(n -> assertEquals(2, users.count(), "UnexpectedUser Amount"));
+    }
+
+    @RepeatedTest(500)
+    public void connectionConcurrencyTest() {
+        AssertSubscriber<String> subscriber = Multi.createFrom().range(0, CONCURRENCY_LEVEL).flatMap(n -> Multi
+                .createFrom().ticks()
+                .every(Duration.ofMillis(getIdleMs() + 3))
+                .onOverflow().drop()
+                .onFailure().invoke(e -> Assertions.fail("Unexpected exception " + e.getMessage()))
+                .onItem().transform(i -> makeAgroalRawQuery()))
+                .subscribe()
+                .withSubscriber(AssertSubscriber.create(CONCURRENCY_LEVEL));
+
+        subscriber
+                .awaitItems(CONCURRENCY_LEVEL)
+                .getItems()
+                .forEach(currentTime -> assertFalse(currentTime.isEmpty(), "Unexpected Application response"));
+    }
+
+    private long getIdleMs() {
+        int idle = Integer.parseInt(idleSec.replaceAll("[A-Z]", ""));
+        return Duration.ofSeconds(idle).toMillis();
+    }
+
+    private Multi<Long> activeConnectionsAsync(int events) {
+        return Multi.createFrom().range(0, events).onItem().transform(i -> activeConnections());
+    }
+
+    private Multi<Long> makeApplicationQueryAsync(int events) {
+        return Multi.createFrom().range(0, events).onItem().transform(i -> {
+            makeApplicationQuery();
+            return 0L;
+        });
+    }
+
+    private Long activeConnections() {
+        Query query = em.createNativeQuery("select * from INFORMATION_SCHEMA.PROCESSLIST;");
+        return (long) query.getResultList().size();
+    }
+
+    private int makeApplicationQuery() {
+        return given()
+                .accept("application/hal+json")
+                .when().get("/users/all")
+                .then()
+                .statusCode(HttpStatus.SC_OK)
+                .extract().response().statusCode();
+    }
+
+    private String makeAgroalRawQuery() {
+        String currentTime = "";
+        try {
+            Connection con = agroalDataSource.getConnection();
+            ResultSet rs = con.createStatement().executeQuery("SELECT CURRENT_TIMESTAMP");
+            rs.next();
+            currentTime = rs.getString(1);
+            rs.close();
+            con.close();
+        } catch (SQLException e) {
+            assertNull(e.getCause(), "makeAgroalRawQuery: Agroal datasource/poolImpl unexpected error");
+        }
+
+        return currentTime;
+    }
+}
--- a/014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/containers/MySqlDatabaseTestResource.java
+++ b/014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/containers/MySqlDatabaseTestResource.java
@@ -25,7 +25,8 @@ public class MySqlDatabaseTestResource implements QuarkusTestResourceLifecycleMa
 
     @Override
     public Map<String, String> start() {
-        container = new MySQLContainer<>(DockerImageName.parse("mysql/mysql-server:8.0.22").asCompatibleSubstituteFor(MYSQL));
+        container = new MySQLContainer<>(
+                DockerImageName.parse("quay.io/bitnami/mysql:5.7.32").asCompatibleSubstituteFor(MYSQL));
         container.withUrlParam("currentSchema", DEFAULT_SCHEMA);
         container.start();
 
--- a/014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/rest/data/UserResourceTest.java
+++ b/014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/rest/data/UserResourceTest.java
@@ -4,9 +4,9 @@ import static io.restassured.RestAssured.given;
 import static org.hamcrest.CoreMatchers.containsString;
 import static org.hamcrest.CoreMatchers.notNullValue;
 import static org.hamcrest.core.IsNot.not;
+import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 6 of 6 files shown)


014-quarkus-panache-with-transactions-xa/src/main/resources/application.properties
```diff
@@ -7,7 +7,13 @@ quarkus.datasource.username=sarah
 quarkus.datasource.password=connor
 quarkus.datasource.jdbc.url=jdbc:mysql://localhost:3306/db?currentSchema=test
 quarkus.datasource.jdbc.transactions=enabled
-quarkus.datasource.jdbc.max-size=1
```
Sgitario, line 10: Nack: Setting `max-size` to 1 was done on purpose to ensure no more than 1 connection is created. Indeed, thanks to this setting, we have been able to spot issues on Quarkus upstream.

jsmrcka:
Got the same fail locally as in GH checks:
```
org.opentest4j.AssertionFailedError: Expected 1 active connections ==> expected: <1> but was: <3>
	at io.quarkus.qe.dbpool.AgroalPoolTest.agroalCheckIdleTimeout(AgroalPoolTest.java:82)
```

jsmrcka:
I'm missing a mention in README about the new test. Some kind of description of does the test verify.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
```diff
@@ -0,0 +1,148 @@
...
+                max.set(activeCon.intValue());
+            }
+            assertTrue(datasourceMaxSize >= activeCon, "More mysql SQL connections than pool max-size");
+            assertTrue(datasourceMinSize <= activeCon, "Less mysql SQL connections than pool min-size");
+            completed.countDown();
+        });
+
+        LOG.info("Max connection requested: " + max);
+        completed.await(TIMEOUT_SEC, TimeUnit.SECONDS);
+        assertEquals(0, completed.getCount(), String.format("Missing %d events.", events - completed.getCount()));
```
jsmrcka, line 88: I find this misleading. `completed.getCount()` returns number of unfinished events, which is by itself a little bit confusing naming. And here, `events - completed.getCount()` returns number of finished events, opposite to what the message states.

This is also the case in `agroalCheckBorderConditionBetweenIdleAndGetConnection`

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    @RepeatedTest(10)
+    @Order(3)
+    @Disabled("https://github.com/quarkusio/quarkus/issues/17141")
+    public void agroalCheckBorderConditionBetweenIdleAndGetConnection() throws InterruptedException {
+        final int concurrentRequestAmount = 2;
+        CountDownLatch completed = new CountDownLatch(concurrentRequestAmount);
+        LOG.info("######################");
```
jsmrcka, line 97: Consider adding a prefix to set it apart from identical log entry in `agroalConcurrentLoadTest`.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
```diff
@@ -0,0 +1,148 @@
...
+        Multi.createFrom().range(0, concurrentRequestAmount).subscribe()
+                .with(n -> Multi.createFrom().ticks().every(Duration.ofSeconds(getIdle() + 3))
+                        .onOverflow().drop()
+                        .onFailure().invoke(e -> Assertions.fail("Unexpected exception " + e.getMessage()))
+                        .subscribe().with(i -> {
+                            LOG.info("Result : " + users.count());
```
jsmrcka, line 103: Consider adding a prefix to set it apart from identical log entry in `agroalConcurrentLoadTest`.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
pjgg: good catch!

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
```diff
@@ -0,0 +1,190 @@
...
+import io.quarkus.qe.rest.data.UserRepository;
+import io.quarkus.test.junit.QuarkusTest;
+import io.quarkus.test.junit.TestProfile;
+import io.smallrye.mutiny.Multi;
+import io.smallrye.mutiny.tuples.Tuple2;
+
+@QuarkusTest
+@TestProfile(MysqlTestProfile.class)
+@TestMethodOrder(OrderAnnotation.class)
+public class AgroalPoolTest {
```
jsmrcka, line 42: The class name already declares that it is about Agroal. So I would drop the `agroal` prefix in test method names.
Also, it would be nice to name all the test methods uniformly, either `testSomething`, `checkSomething` or just `something`. Now it is a mixture.
WDYT?

jsmrcka (changes requested):

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
```diff
@@ -0,0 +1,188 @@
...
+                .onOverflow().drop()
+                .onFailure().invoke(e -> Assertions.fail("Unexpected exception " + e.getMessage()))
+                .subscribe().with(i -> {
+                    makeApplicationQuery();
+                    queuedEvents.countDown();
+                }));
+
+        queuedEvents.await(TIMEOUT_SEC, TimeUnit.SECONDS);
+        assertEquals(0, queuedEvents.getCount(),
+                String.format("agroalCheckBorderConditionBetweenIdleAndGetConnection: Missing %d events.", queuedEvents.getCount()));
```
jsmrcka, line 116: I think the formatting check is failing on this line being too long.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
```diff
@@ -0,0 +1,188 @@
...
+                .every(Duration.ofSeconds(getIdle() + 3))
+                .onOverflow().drop()
+                .onFailure().invoke(e -> Assertions.fail("agroalConnectionConcurrency: Unexpected exception " + e.getMessage()))
+                .subscribe().with(i -> {
+                    makeAgroalRawQuery();
+                    queuedEvents.countDown();
+                }));
+
+        queuedEvents.await(TIMEOUT_SEC, TimeUnit.SECONDS);
+        assertEquals(0, queuedEvents.getCount(), String.format("agroalConnectionConcurrency: Missing %d events.", queuedEvents.getCount()));
```
jsmrcka, line 142: I think the formatting check is failing on this line being too long.

jsmrcka (approved):
LGTM. Thanks :).

014-quarkus-panache-with-transactions-xa/README.md
```diff
@@ -15,4 +15,6 @@ Additional tests:
 Additional UserEntity is a simple JPA entity that was created with aim to avoid inheritance of PanacheEntity methods
 and instead test the additional combination of JPA entity + PanacheRepository + PanacheRepositoryResource, where
 PanacheRepository is a facade class. Facade class can override certain methods to change the default behaviour of the
-PanacheRepositoryResource methods.
\ No newline at end of file
+PanacheRepositoryResource methods.
+
+- AgroalPoolTest, will cover how the db pool is manage in terms of IDLE-timeout, max connections and concurrency.
```
Sgitario, line 20:
```suggestion
- AgroalPoolTest will cover how the db pool is manage in terms of IDLE-timeout, max connections and concurrency. 
```

Sgitario, line 20:
```suggestion
- AgroalPoolTest will cover how the db pool is managed in terms of IDLE-timeout, max connections and concurrency. 
```

014-quarkus-panache-with-transactions-xa/src/main/resources/application.properties
```diff
@@ -16,4 +16,19 @@ quarkus.datasource.with-xa.jdbc.url=jdbc:mysql://localhost:3306/db?currentSchema
 quarkus.datasource.with-xa.jdbc.transactions=xa
 
 quarkus.hibernate-orm.database.generation=drop-and-create
-quarkus.hibernate-orm.sql-load-script=import.sql
\ No newline at end of file
+quarkus.hibernate-orm.sql-load-script=import.sql
+
+
```
Sgitario, line 21: Two emtpy lines here.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/containers/MysqlTestProfile.java
```diff
@@ -0,0 +1,21 @@
+package io.quarkus.qe.containers;
+
+import java.util.Collections;
+import java.util.List;
+
+import io.quarkus.test.junit.QuarkusTestProfile;
+
+public class MysqlTestProfile implements QuarkusTestProfile {
```
Sgitario, line 8: QuarkusTestProfile is not compatible when running the test on Native, so we're missing this coverage on Native. Do you think this coverage is not important for Native, so we can live without it?

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
```diff
@@ -0,0 +1,188 @@
...
+import io.quarkus.qe.containers.MysqlTestProfile;
+import io.quarkus.qe.rest.data.UserRepository;
+import io.quarkus.test.junit.QuarkusTest;
+import io.quarkus.test.junit.TestProfile;
+import io.smallrye.mutiny.Multi;
+import io.smallrye.mutiny.tuples.Tuple2;
+
+@QuarkusTest
+@TestProfile(MysqlTestProfile.class)
+@TestMethodOrder(OrderAnnotation.class)
```
Sgitario, line 41: Is the test method order a requirement to run this test? Having tests that depend on each other is most of the times a bad practice.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
```diff
@@ -0,0 +1,188 @@
...
+    int datasourceMaxSize;
+
+    @ConfigProperty(name = "quarkus.datasource.jdbc.min-size")
+    int datasourceMinSize;
+
+    @Test
+    @Order(1)
+    public void idleTimeoutTest() throws InterruptedException {
+        makeApplicationQuery();
+        Thread.sleep(3000);
```
Sgitario, line 70: Why 3 seconds? If this value depends on a property, I would rather using this property instead of a fixed value.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
```diff
@@ -0,0 +1,188 @@
...
+        assertEquals(1, activeConnections(), "agroalCheckIdleTimeout: Expected " + datasourceMinSize + " active connections");
+    }
+
+    @Test
+    @Order(2)
+    public void poolTurnoverTest() throws InterruptedException {
+        final int events = 2500;
```
Sgitario, line 77: I think `final` is redundant.
Also, why 2500 events? It seems to me like too much.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
```diff
@@ -0,0 +1,188 @@
...
+        final AtomicInteger max = new AtomicInteger(0);
```
Sgitario, line 78: I think `final` is redundant.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
```diff
@@ -0,0 +1,188 @@
...
+        LOG.info("agroalCheckDbPoolTurnover: Max connection requested: " + max);
+        queuedEvents.await(TIMEOUT_SEC, TimeUnit.SECONDS);
+        assertEquals(0, queuedEvents.getCount(),
+                String.format("agroalCheckDbPoolTurnover: Missing %d events.", queuedEvents.getCount()));
+    }
+
+    @RepeatedTest(10)
+    @Order(3)
+    public void borderConditionBetweenIdleAndGetConnectionTest() throws InterruptedException {
+        final int concurrentRequestAmount = 2;
```
Sgitario, line 103: I think `final` is redundant.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
```diff
@@ -0,0 +1,188 @@
...
+        CountDownLatch queuedEvents = new CountDownLatch(concurrentRequestAmount);
+        Multi.createFrom().range(0, concurrentRequestAmount).subscribe().with(n -> Multi.createFrom().ticks()
+                .every(Duration.ofSeconds(getIdle() + 3))
+                .onOverflow().drop()
+                .onFailure().invoke(e -> Assertions.fail("Unexpected exception " + e.getMessage()))
+                .subscribe().with(i -> {
+                    makeApplicationQuery();
+                    queuedEvents.countDown();
+                }));
```
Sgitario, line 112: I don't understand very well this code.
Are you running two concurrent threads and each thread is making calls to the application every 3 seconds? Then, we verify that the queuedEvents is 0... 
Why do we need 2 threads? If one of them fails, the another thread can call the application twice and then the test will still satisfy the condition.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
```diff
@@ -0,0 +1,188 @@
...
+        assertEquals(0, queuedEvents.getCount(),
+                String.format("agroalCheckBorderConditionBetweenIdleAndGetConnection: Missing %d events.", queuedEvents.getCount()));
+    }
+
+    @RepeatedTest(100)
+    @Order(4)
+    public void concurrentLoadTest() {
+        final int concurrentRequestAmount = 20;
+        Multi.createFrom().range(0, concurrentRequestAmount).subscribe()
+                .with(n -> LOG.info("agroalConcurrentLoadTest: Result : " + users.count()));
```
Sgitario, line 124: This test depends on the execution of other tests which is generally a bad practice. Can we assert this condition as part of the other code instead?

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    @RepeatedTest(250)
+    @Order(5)
+    public void connectionConcurrencyTest() throws SQLException, InterruptedException {
+        final int concurrentRequestAmount = 2;
+        CountDownLatch queuedEvents = new CountDownLatch(concurrentRequestAmount);
+        Multi.createFrom().range(0, concurrentRequestAmount).subscribe().with(n -> Multi.createFrom().ticks()
```
Sgitario, line 132: Same comment as in the method borderConditionBetweenIdleAndGetConnectionTest

Sgitario (changes requested):
Apart from the comments, the most important topics are:
- We're missing the Native coverage and from my point of view this would be necessary as well.
- Why are you using the `@RepeatedTest` annotation? Repeating the test too many times seems too much for me.
- Tests depend on the method order to work fine and generally speaking this is a bad practice.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/containers/MysqlTestProfile.java
pjgg: is not intended to be a native test.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
pjgg: Not really is pending to a Zulip resolution. I have a question related to `quarkus.datasource.jdbc.idle-removal-interval`. But I agree we can disable this test and remove the order stuff until having a reply.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
pjgg: is not every 3 sec. We are making a query 3 ms ahead of the IDLE-timeout. 

`.every(Duration.ofSeconds(getIdle() + 3))`

We are checking border condition errors.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
pjgg: All tests are an isolated tests.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
Sgitario: I still don't understand this method. As I can read: `.every(Duration.ofSeconds(1 + 3))`, It's traslated to every 4 seconds. If it's for some reason, it's indeed 3 or 4 ms, can you make it more readable please?

Sgitario (changes requested):
Some comments are marked as resolved, but they're not.
And this comment https://github.com/quarkus-qe/beefy-scenarios/pull/203/files#r632530612 is still unclear to me.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
pjgg: Ok I have changed a little bit the signature in order to be more readable:

`.every(Duration.ofSeconds(getIdleMs() + 3))`

- 2 concurrent threads and then making a request every X time: 

concurrent queries every X secs N times (Repeat annotation) is exactly the aim of this test. This is why is called `connectionConcurrencyTest`, or in other words, check that two (or more) concurrent queries completed successfully. In case of any issue, ` queuedEvents.countDown()` is not going to be invoked and the test is going to fail 

```
assertEquals(0, queuedEvents.getCount(),
                String.format("agroalConnectionConcurrency: Missing %d events.", queuedEvents.getCount()));
```

 - If what we want is to stress the db pool, what about running this test using more than 2 threads (which I think it simulates the users):
 
  we could increase the number of concurrent queries (I did , is more expensive in terms of CPU and time, I have increased from 2 to 20) but, the `stress` feature is cover by the `RepeatedTest` annotation (is a combination of both). So, 2 concurrent queries repeated 500 times, means 1000 consecutive successful queries. 

With the current concurrent configuration, attending to your request, we are going to make 10k consecutive successful queries.  Is going to be more expensive(10 times), I thought that 1000 queries were enough to probe that we handler concurrency as expected ( some issues have risen with even less iteration and concurrency), from a agroal datasource point of view, but is also ok)

pjgg:
> Some comments are marked as resolved, but they're not.
> And this comment https://github.com/quarkus-qe/beefy-scenarios/pull/203/files#r632530612 is still unclear to me.

Which comment si is not clear? 

this one? This test depends on the execution of other tests which is generally a bad practice. Can we assert this condition as part of the other code instead?

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
Sgitario: I see you have changed the every method to:
`.every(Duration.ofSeconds(getIdleMs() + 3))`

But it's still odd to me as you have milliseconds and use it as seconds.
Let me be clearer, this is what you are doing:

```
int concurrentRequestAmount = 2; // To mimic users
int idle = 1; // 1 seconds
long idleMs = Duration.ofSeconds(idle).toMillis(); // 1000 milliseconds
CountDownLatch queuedEvents = new CountDownLatch(concurrentRequestAmount); // Wait for 2 requests
Multi.createFrom().range(0, concurrentRequestAmount).subscribe()
        .with(n -> Multi.createFrom().ticks()
                .every(Duration.ofSeconds(idleMs + 3)) // 1000 + 3 as seconds 
                .onOverflow().drop()
                .subscribe().with(i -> {
                    System.out.println("Tick: " + i);
                    queuedEvents.countDown();
                }));
assertTrue(queuedEvents.await(5, TimeUnit.SECONDS));     
```

This is working for you, because you're waiting for 2 requests done by 2 concurrent requests, but the `Multi.createFrom().ticks().every(Duration.ofSeconds(idleMs + 3))` is doing a request every 1003 seconds now (before it was 4 seconds).

I think the right code is to change it to `Multi.createFrom().ticks().every(Duration.ofMillis(idleMs + 3))`.

However, as you are verifying that the requests of the concurrent users worked fine, the current solution would be the same as doing:

```
final int concurrentRequestAmount = 20;
CountDownLatch queuedEvents = new CountDownLatch(concurrentRequestAmount);
Multi.createFrom().range(0, concurrentRequestAmount).subscribe().with(n -> {
            makeApplicationQuery();
            queuedEvents.countDown();
        });
```

But guessing that what you exactly want to stress the border condition, then another idea would be to either assert there are no failures after some amount of time or verify the requests after some ticks (`new CountDownLatch(concurrentRequestAmount * NUM_OF_ITERATIONS or ticks);`

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
```diff
@@ -0,0 +1,188 @@
...
+    int datasourceMaxSize;
+
+    @ConfigProperty(name = "quarkus.datasource.jdbc.min-size")
+    int datasourceMinSize;
+
+    @Test
+    @Disabled("Pending Zulip conversation about quarkus.datasource.jdbc.idle-removal-interval")
+    public void idleTimeoutTest() throws InterruptedException {
+        makeApplicationQuery();
+        Thread.sleep(Duration.ofSeconds(getIdleMs() + 1).toMillis());
```
Sgitario, line 71: Is this working? You are parsing 1000 milliseconds + 1, then converting it to seconds, which is 1001 seconds and then to milliseconds 1001000. So, you're sleeping the test for 1001000 milliseconds.

Sgitario (changes requested):

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
pjgg: Was leftover as a result of a side effect from your previous comments.

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/dbpool/AgroalPoolTest.java
pjgg: Yes in this test `borderConditionBetweenIdleAndGetConnectionTest` the key concept is the IDLE-timeout border condition.

Sgitario (approved):

Sgitario:
Waiting for CI before merging.

Sgitario (changes requested):
The CI action failed for Native. I'm aware that the changes do no affect the Native build, but I think it spotted an issue as the AgroalPoolTest test was stucked. From logs:

```
2021-05-17T09:09:46.9832375Z 2021-05-17 09:09:42,758 WARN  [io.agr.pool] (executor-thread-39) Datasource '<default>': JDBC resources leaked: 0 ResultSet(s) and 1 Statement(s)
2021-05-17T09:09:46.9833909Z 2021-05-17 09:09:42,759 INFO  [io.qua.qe.dbp.AgroalPoolTest] (executor-thread-103) 2021-05-17 09:09:42
2021-05-17T09:09:46.9835516Z 2021-05-17 09:09:42,759 WARN  [io.agr.pool] (executor-thread-103) Datasource '<default>': JDBC resources leaked: 0 ResultSet(s) and 1 Statement(s)
2021-05-17T09:09:46.9838630Z 2021-05-17 09:09:42,759 INFO  [io.qua.qe.dbp.AgroalPoolTest] (executor-thread-112) 2021-05-17 09:09:42
2021-05-17T09:09:46.9843561Z 2021-05-17 09:09:42,759 WARN  [io.agr.pool] (executor-thread-112) Datasource '<default>': JDBC resources leaked: 0 ResultSet(s) and 1 Statement(s)
2021-05-17T09:09:46.9847045Z 2021-05-17 09:09:42,759 INFO  [io.qua.qe.dbp.AgroalPoolTest] (executor-thread-112) 2021-05-17 09:09:42
```

Sgitario (approved):

PR #207, opened 2021-05-17
Add coverage for `@ConfigMapping` annotation

Sgitario:
Examples are inspired by https://github.com/smallrye/smallrye-config/blob/main/cdi/src/test/java/io/smallrye/config/inject/ConfigMappingInjectionTest.java
More info about this annotation: https://smallrye.io/docs/smallrye-config/mapping/mapping.html

### DIFF
--- a/022-quarkus-properties-config-all/src/main/java/io/quarkus/qe/configmapping/Person.java
+++ b/022-quarkus-properties-config-all/src/main/java/io/quarkus/qe/configmapping/Person.java
@@ -1,0 +1,9 @@
+package io.quarkus.qe.configmapping;
+
+import io.smallrye.config.ConfigMapping;
+
+@ConfigMapping(prefix = "person")
+public class Person {
+    String name;
+    int age;
+}
--- a/022-quarkus-properties-config-all/README.md
+++ b/022-quarkus-properties-config-all/README.md
@@ -24,6 +24,8 @@ supported classes (e.x. String, Boolean, Double, etc.)
 
 6. Injecting a `ConfigValue` instance that was introduced in MicroProfile Config 2.0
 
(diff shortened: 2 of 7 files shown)


Sgitario:
@kshpak would you mind to take a look into this PR? Thanks in advance,

kshpak:
I will take a look in 30 mins

kshpak:
The test is failing with latest quarkus main. Apparently, io.smallrye.config expect that all properties with prefix "server" will be used

Sgitario:
> The test is failing with latest quarkus main. Apparently, io.smallrye.config expect that all properties with prefix "server" will be used

Fixed. Would you mind taking another look?

022-quarkus-properties-config-all/src/main/java/io/quarkus/qe/configmapping/Person.java
```diff
@@ -0,0 +1,17 @@
...
+    int age;
+
+    public String getName() {
+        return name;
+    }
+
+    public int getAge() {
+        return age;
+    }
+}
```
kshpak, line 17: getters can be deleted and a field can be accepted directly by calling a variable name. IMO this is the added value of config class vs. config interface

kshpak: Now I realize that the additional logic can be placed under getters. This is up to you now.

kshpak (approved):

022-quarkus-properties-config-all/src/main/java/io/quarkus/qe/configmapping/Person.java
Sgitario: Updated

kshpak (approved):

PR #208, opened 2021-05-17
Move on 602-spring-data-rest module to devServices

pjgg:
- DevService is a new feature that allows you to use a default / autogenerated data source on Dev and Test mode. This PR, migrate an existing module that uses a hardcode datasource to devServices.


Sgitario (changes requested):
I've just checked that the testcontainers needs to be removed too: the testcontainer:postgresql dependency and the testresources used in the tests.

602-spring-data-rest/pom.xml
```diff
@@ -16,10 +16,6 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-resteasy-jackson</artifactId>
         </dependency>
-        <dependency>
-            <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-jdbc-postgresql</artifactId>
-        </dependency>
```
Sgitario, line 22: The testcontainers dependency, not the quarkus jdbc dependency.

Sgitario (changes requested):

Sgitario (changes requested):
Pablo, testcontainers and the test resources that are used in the tests of this module.

Sgitario (approved):

PR #215, opened 2021-05-24
Vertx Logger and ConfigProperties was deprecated

pjgg:
Upgrade deprecated code: 

`io.quarkus.arc.config.ConfigProperties` was deprecated
`Vertx` Logger was deprecated


pjgg:
Looks like `io.smallrye.config.ConfigMapping` doesn't behaves exactly the same as the deprecated `io.quarkus.arc.config.ConfigProperties`. 

ConfigProperties is able to read child properties from a parent class configuration:
Example:

```
@ConfigProperties(prefix = "antagonist")
public class AntagonistConfiguration {
    public String message;
    public String name;
    public AntagonistWifeConfig wife;

    public static class AntagonistWifeConfig {
        public String name;
        public String message;
    }
}
```

application.properties

```
antagonist.name=Sheldon Plankton
antagonist.message=Hi, I am Sheldon Plankton
antagonist.wife.name=Karen
antagonist.wife.message=Hi, I am Karen
```

But the same scenario with `ConfigMapping` throw an exception

```
SRCFG00013: No Converter registered for class io.quarkus.qe.config.AntagonistConfiguration$AntagonistWifeConfig
```

rsvoboda:
@pjgg this is candidate for a bug report. The transition should be smooth for the users.

010-quarkus-opentracing-reactive-grpc/src/test/java/io/quarkus/qe/GrpcPingPongResourceTest.java
```diff
@@ -14,6 +15,7 @@
 public class GrpcPingPongResourceTest extends AbstractPingPongResourceTest {
 
     @Inject
+    @GrpcService
```
jsmrcka, line 18: This has already been merged into `main`, so the diff should vanish after rebase.

jsmrcka (commented):
What is the plan with the different behavior of `@ConfigMapping`? (See https://github.com/quarkusio/quarkus/issues/17545#issuecomment-851580648, which corresponds with https://smallrye.io/docs/smallrye-config/main/mapping/mapping.html.)

I see no way to smoothly transition from the deprecated annotation.

pjgg:
> What is the plan with the different behavior of `@ConfigMapping`? (See [quarkusio/quarkus#17545 (comment)](https://github.com/quarkusio/quarkus/issues/17545#issuecomment-851580648), which corresponds with https://smallrye.io/docs/smallrye-config/main/mapping/mapping.html, also referenced in https://quarkus.io/version/main/guides/config#inject-the-configuration.)
> 
> I see no way to smoothly transition from the deprecated annotation.

The new implementation is great!. To me makes even more sense than the old one from a conceptual point of view. 

For example the above code will look something like this:

```
@ConfigMapping(prefix = "antagonist")
public class AntagonistConfiguration {
    public String message;
    public String name;
    public AntagonistWifeConfig wife;

    public interface AntagonistWifeConfig {
        String name();
        String message();
    }
}
```

But the point is that this change is a breaking change that will makes fail all the existing applications that use this feature. So, we should think about a smooth transition.

PR #217, opened 2021-06-03
Disable Native tests for module 022  because #quarkus/17653

Sgitario:
Adding external properties on Native make the app fail to start. Reported by https://github.com/quarkusio/quarkus/issues/17653

### DIFF
--- a/022-quarkus-properties-config-all/pom.xml
+++ b/022-quarkus-properties-config-all/pom.xml
@@ -20,6 +20,11 @@
     <profiles>
         <profile>
             <id>native</id>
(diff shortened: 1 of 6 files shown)


kshpak:
Will take a look tommorow

kshpak:
@Sgitario the problem not in converter, but in new library introduced by this commit. I'm working on changes. ConfigProperties annotation should be used instead of the `io.smallrye.config.ConfigMapping`. But it throws errors, which I'm trying to fix

kshpak (approved):

PR #220, opened 2021-06-07
Disable module 002-quarkus-all-extensions

pjgg:
The missing productized version of "scala-compiler" in artifact repository  make module fail


pjgg:
Let's close this ticket, should be fixed

PR #223, opened 2021-06-08
Fix 009 module to use @Identifier instead of @Named to inject broker

Sgitario:
This change is related to https://github.com/quarkusio/quarkus/commit/11bc00f324b6ed4166ab4250e6661060b7d1a4d4.
The problem is that keeping injecting the default Kafka broker using `@Named` is now failing:

```java
@ApplicationScoped
public class KafkaProviders {

    @Inject
    @Named("default-kafka-broker")
    Map<String, Object> config;
}
```

Using:

```java
@ApplicationScoped
public class KafkaProviders {

    @Inject
    @Identifier("default-kafka-broker")
    Map<String, Object> config;
}
```

It works. 

Therefore, this is a breaking change that should be noted in the Migration guide for 2.x. Reported by https://github.com/quarkusio/quarkus/issues/17754


Sgitario:
This is making Beefy TS fail. Merging. @pjgg, If you see something to address, feel free to create a separate PR.

PR #228, opened 2021-06-09
Disable Kamelet module

pjgg:
Quarkus Kamelet doesn't exist in Quarkus Kamelet 1.6

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,7 +29,7 @@
       <module>019-quarkus-consul</module> -->
       <module>020-quarkus-http-non-application-endpoints</module>
       <module>021-quarkus-panache-multiple-pus</module>
-      <module>023-quarkus-kamelet</module>
+      <!-- <module>023-quarkus-kamelet</module> -->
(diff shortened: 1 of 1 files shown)


rsvoboda (commented):
How do we ensure https://issues.redhat.com/browse/QUARKUS-1039 when this module gets diasbled?

Sgitario (changes requested):
Please, amend also the version: https://github.com/quarkus-qe/beefy-scenarios/blob/752f18150b5c4da6ff2dfa4f0f79bf594a1f737e/pom.xml#L56.
It should be 1.6.0.

Sgitario (commented):
This PR looks good to me now. 
However, as far I see, the above comment from @rsvoboda needs to be clarified/addressed before merging.

rsvoboda (approved):
Approving, QUARKUS-1039 can be checkd for code change presence, Pablo will also try to look at another approach without kamelet.

Sgitario (approved):

PR #232, opened 2021-06-16
[WIP] Kafka-client SSL scenario

pjgg:
SSL Kafka-client test coverage


009-quarkus-infinispan-grpc-kafka/src/main/java/io/quarkus/qe/providers/KafkaProviders.java
```diff
@@ -1,52 +1,84 @@
...
 @ApplicationScoped
 public class KafkaProviders {
 
-    @Inject
-    @Identifier("default-kafka-broker")
-    Map<String, Object> config;
+    @ConfigProperty(name = "kafka-client.bootstrap.servers", defaultValue = "localhost:9092")
+    String kafkaBootStrap;
+
+    @ConfigProperty(name = "ssl-dir", defaultValue = "src/test/resources")
```
Sgitario, line 28: The default value is tricky. Main sources should not rely on test sources.

Sgitario (commented):
- What part of Quarkus are we covering in here? I see you have configured Kafka with SSL, but didn't use anything in Quarkus. I thought this coverage was about to use the properties around `quarkus.kafka-streams.ssl` which is already covered in upstream. So I don't know what's the benefit.
- Losing the normal behaviour (without SSL): I would have expected to keep the non SSL and SSL scenarios at the same time.

I need some clarifications before proceeding with the review.

pjgg:
> * What part of Quarkus are we covering in here? I see you have configured Kafka with SSL, but didn't use anything in Quarkus. I thought this coverage was about to use the properties around `quarkus.kafka-streams.ssl` which is already covered in upstream. So I don't know what's the benefit.
> * Losing the normal behaviour (without SSL): I would have expected to keep the non SSL and SSL scenarios at the same time.
> 
> I need some clarifications before proceeding with the review.

This PR cover Quarkus Kafka-client over a Kafka with SSL.

pjgg:
moved to #233

PR #235, opened 2021-06-18
Wait for Jaeger container to be ready

kshpak:
These 3 tests were constantly failing because of the Jaeger collector docker container


rsvoboda (approved):
Changes look good to me, @pjgg waiting for your feedback.

@kshpak so 30s timeout was not enough? Do you know how the "boot" times changed?

kshpak:
> Changes look good to me, @pjgg waiting for your feedback.
> 
> @kshpak so 30s timeout was not enough? Do you know how the "boot" times changed?

The problem is in function, withStartupTimeout is not the right choice for that, it's only waiting for some ports to be ready. It evaluates readiness too early, some services didn't load at that time.

PR #240, opened 2021-07-08
Panache sort is already implemented

pjgg:
Remove workaround.
Feature Implemented: https://github.com/quarkusio/quarkus/issues/16871

### DIFF
--- a/014-quarkus-panache-with-transactions-xa/src/main/java/io/quarkus/qe/rest/data/UserRepository.java
+++ b/014-quarkus-panache-with-transactions-xa/src/main/java/io/quarkus/qe/rest/data/UserRepository.java
@@ -1,18 +1,16 @@
 package io.quarkus.qe.rest.data;
 
+import io.quarkus.hibernate.orm.panache.PanacheQuery;
+import io.quarkus.panache.common.Sort;
 import javax.enterprise.context.ApplicationScoped;
 
-import io.quarkus.hibernate.orm.panache.PanacheQuery;
 import io.quarkus.hibernate.orm.panache.PanacheRepositoryBase;
-import io.quarkus.panache.common.Sort;
 
 @ApplicationScoped
 public class UserRepository implements PanacheRepositoryBase<UserEntity, Long> {
 
-    //This will override behaviour of HTTP GET method of PanacheResource
-    //TODO change to findAll() once feature is implemented https://github.com/quarkusio/quarkus/issues/16871
     @Override
-    public PanacheQuery<UserEntity> findAll(Sort sort) {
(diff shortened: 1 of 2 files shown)


014-quarkus-panache-with-transactions-xa/src/main/java/io/quarkus/qe/rest/data/UserRepository.java
```diff
@@ -2,17 +2,8 @@
...
 
 @ApplicationScoped
 public class UserRepository implements PanacheRepositoryBase<UserEntity, Long> {
-
-    //This will override behaviour of HTTP GET method of PanacheResource
-    //TODO change to findAll() once feature is implemented https://github.com/quarkusio/quarkus/issues/16871
-    @Override
-    public PanacheQuery<UserEntity> findAll(Sort sort) {
-        return find("select u from user_entity u order by u.name asc");
-    }
```
kshpak, line 17: This method should remain but be changed to `public PanacheQuery<UserEntity> findAll()`.
The point was to override the default behavior of HTTP GET, without touching HTTP GET?sort=

As I recall it right, if you do `get("/users/all?sort=name")` and `get("/users/all?sort=!name")` the output was the same (because we overrided findAll(Sort sort) to sort ascending)

The fixed issue should verify that by overriding `findAll()` we do override HTTP GET **without** parameters
&& by overriding `findAll(Sort sort)` we override HTTP GET **with** parameters

Before, we could change the behavior of HTTP GET without parameters only by overriding `findAll(Sort sort)`, but this will also affect HTTP GET with parameters (which is undesirable)

014-quarkus-panache-with-transactions-xa/src/test/java/io/quarkus/qe/rest/data/UserResourceTest.java
```diff
@@ -26,7 +26,7 @@ void testAllRepositoryMethods() {
         //Assert that order matches. UserRepository override it to be ascend. Initial order defined by import.sql
         String userList = given()
                 .accept("application/hal+json")
-                .when().get("/users/all")
+                .when().get("/users/all?sort=name")
```
kshpak, line 29: This can stay as is (i.e `when().get("/users/all)`)
The behavior of `get("/users/all?sort=!name")` can be checked, to verify it's sorting in descent order.

kshpak (changes requested):
Please add `public PanacheQuery<UserEntity> findAll()`
&& test descending order is working properly if parameter was specified

kshpak (approved):

PR #242, opened 2021-07-20
Run Native only for the changed modules in PRs

Sgitario:
This solution is based on the Jenkins script detection file and uses the https://github.com/marketplace/actions/get-all-changed-files step to get the changes for a PR.

Furthermore, the PR will address a couple of bugs:
- In 014, it will fix a format issue (we should take care of merging PRs with failing actions)
- In 302, the behaviour of `@WithName` has changed in 2.1. I've reported https://github.com/quarkusio/quarkus/issues/18842 to clarify if the new behaviour is correct (I think it is) and to update the migration guide accordingly.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -33,8 +33,8 @@ jobs:
         with:
           name: maven-repo
           path: maven-repo.tgz
-  linux-build:
-    name: Linux - PR build
+  linux-build-jvm:
+    name: PR - Linux - JVM build
     runs-on: ubuntu-latest
     needs: quarkus-main-build
     strategy:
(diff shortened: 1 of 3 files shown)


.github/workflows/pr.yaml
```diff
@@ -66,20 +66,73 @@ jobs:
...
+        with:
+          name: maven-repo
+          path: .
+      - name: Extract Maven Repo
+        shell: bash
+        run: tar -xzf maven-repo.tgz -C ~
+      - id: files
+        uses: jitterbit/get-changed-files@v1
+      - name: Test in Native mode
+        run: |
```
rsvoboda, line 109: If `pom.xml` is changed we can assume Quarkus version got bumped, so I would add `003-quarkus-many-extensions` by default. Even if additional modules got changed too as part of the PR.

rsvoboda (changes requested):

.github/workflows/pr.yaml
Sgitario: +1 will do

Sgitario: Done

rsvoboda (approved):

PR #243, opened 2021-08-04
Fix 013 failure by updating openapi assert

Sgitario:
This is due to this change: quarkusio/quarkus/pull/19148


Sgitario:
It seems the test is no longer failing. Closing...

PR #248, opened 2021-08-18
Separate @QuarkusTest execution

gsmet:
At some point, maybe this should be put in a shared constant but it
seems a bit too involved to create a shared module just for that.

### DIFF
--- a/012-quarkus-kafka-streams/src/test/java/io/quarkus/qe/SslAlertMonitorTest.java
+++ b/012-quarkus-kafka-streams/src/test/java/io/quarkus/qe/SslAlertMonitorTest.java
@@ -1,10 +1,15 @@
 package io.quarkus.qe;
 
+import org.junit.jupiter.api.Tag;
+
 import io.quarkus.qe.resources.SslStrimziKafkaTestResource;
 import io.quarkus.test.common.QuarkusTestResource;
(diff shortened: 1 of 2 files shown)


gsmet:
This should make CI green again \o/

rsvoboda:
@gsmet why does `QuarkusProdModeTest` need separate surefire execution?

gsmet:
Because you cannot mix them with the regular unit tests. It has always been like that and was enforced at some point but the check was removed somehow.

It got reintroduced here: https://github.com/quarkusio/quarkus/pull/19329 and this comment from Stuart: https://github.com/quarkusio/quarkus/issues/19325#issuecomment-896378222 .

rsvoboda:
Thanks for details!

Sgitario (approved):

rsvoboda (approved):

gsmet:
Now that I think of it a bit more, I think we need to isolate the `@QuarkusTest` not the prod mode as this is the one conflicting.

I'll change the PR and push a PR on core to add a `@Tag` for `@QuarkusTest`.

gsmet:
I adjusted the PR and created https://github.com/quarkusio/quarkus/pull/19477 so that at some point we don't have to tag the tests anymore.

PR #254, opened 2021-09-21
Conflicts between reactive and non-reactive libraries

pjgg:
- Disabled DevServices for all extensions module
- Comment Quarkus-SpringWeb extension due to some reactive library conflicts
- patch some other modules in order to fix dependency conflicts

### DIFF
--- a/002-quarkus-all-extensions/src/main/resources/application.properties
+++ b/002-quarkus-all-extensions/src/main/resources/application.properties
@@ -1,5 +1,6 @@
 # Configuration file
 # key = value
+
 quarkus.devservices.enabled=false
 
 # Flyway minimal config properties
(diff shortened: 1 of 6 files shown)


002-quarkus-all-extensions/src/main/resources/application.properties
```diff
@@ -2,6 +2,8 @@
 # key = value
 quarkus.devservices.enabled=false
 
+quarkus.devservices.enabled=false
```
rsvoboda, line 5: already there, two lines above :)

rsvoboda (changes requested):
just small change needed

pjgg:
The native error is because `002-quarkus-all-extensions` module is not included in the native profile.

rsvoboda (approved):

PR #258, opened 2021-10-15
Fix 022-properties bulk properties test with right annotation

kshpak:


Sgitario (approved):

Sgitario:
Failure is unrelated to this PR, merging.

PR #261, opened 2021-11-04
Add quarkus-resteasy-jackson dependency for 603-spring-web module

kshpak:
Additional dep is needed because of decoupling of RestEasy module from quarkus-spring-web


rsvoboda (approved):

rsvoboda:
Let's wait for CI before merging

PR #263, opened 2021-11-15
Make Quartz scheduling test more robust

jsmrcka:
Two-node cluster test behaved unpredictably, because
a cluster executes a scheduled trigger randomly all over
its nodes. The original test expected only one node being
fired while both were running.

### DIFF
--- a/017-quartz-cluster/src/test/java/io/quarkus/qe/quartz/AnnotationScheduledJobsQuartzTest.java
+++ b/017-quartz-cluster/src/test/java/io/quarkus/qe/quartz/AnnotationScheduledJobsQuartzTest.java
@@ -1,10 +1,12 @@
 package io.quarkus.qe.quartz;
 
 import static io.restassured.RestAssured.get;
-import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.time.Duration;
+import java.util.Arrays;
+import java.util.List;
 
 import org.awaitility.Awaitility;
 import org.junit.jupiter.api.BeforeAll;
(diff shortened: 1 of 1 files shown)


017-quartz-cluster/src/test/java/io/quarkus/qe/quartz/AnnotationScheduledJobsQuartzTest.java
```diff
@@ -57,15 +64,23 @@ private void whenShutdownNodeOne() {
...
+    private void thenJobIsExecutedWithOwner(String... expectedOwners) {
         RestAssured.port = REST_PORT;
 
         Awaitility.await().atMost(Duration.ofSeconds(30)).untilAsserted(() -> {
             ExecutionEntity[] executions = get("/executions")
                     .then().statusCode(200)
                     .extract().as(ExecutionEntity[].class);
-
-            assertEquals(expectedOwner, executions[executions.length - 1].owner, "Expected owner not found");
+            assertTrue(Arrays.asList(expectedOwners).contains(executions[executions.length - 1].owner),
```
rsvoboda, line 82: You are checking just one owner, I'd rather see something like

```java
            Arrays.stream(expectedOwners).forEach(expectedOwner -> {
                boolean expectedOwnerFound = Arrays.stream(executions)
                        .anyMatch(execution -> execution.owner.equals(expectedOwner));
                assertTrue(expectedOwnerFound, "Expected owner not found");
            });
```

rsvoboda (changes requested):

017-quartz-cluster/src/test/java/io/quarkus/qe/quartz/AnnotationScheduledJobsQuartzTest.java
jsmrcka: The idea is to check the last one. The previous ones can be anything, really, because the list contains the entire history since DB init. Only the last item in the list (after some time) shows which nodes are currently active.
The suggestion would mostly work, but I think it would not reflect the purpose of the test.

rsvoboda: backlog from chat

Rostislav Svoboda, 13 min
Reading your comment
let's imagine case          thenJobIsExecutedWithOwner(NODE_ONE_NAME, NODE_TWO_NAME);
you are checking if the list o expectedOwners contains executions[executions.length - 1].owner
so you are checking for just one element
I would say that both expectedOwners should be checked for presence in executions.owner
Rostislav Svoboda, 9 min, Edited
Now you are checking that the last executions.owner is in the list of expectedOwners
I would expect the check to be done vice versa, to ensure all the expectedOwners are present

Josef Smrcka, 6 min
As I understand the test, the goal is to check just the last item in executions, which is also the most recently recorded one.
Basically, we don't care about the previous ones, because they're history.

Rostislav Svoboda, 3 min
So pls adjust the message in the assertion to give more context
Josef Smrcka, 1 min
Okay.

Rostislav Svoboda, 1 min
something like "Owner '" + executions[executions.length - 1].owner + "' not found in the list of possible owners " + expectedOwners
and maybe change the name from expectedOwners to possibleOwners
expectedOwners indicate to me that all of them need to be present

rsvoboda (approved):

PR #264, opened 2021-11-17
Comment `303-quarkus-vertx-sql` and `304-quarkus-vertx-routes` due to an issue on hibernate validator

pjgg:
Native compilation error:

```
`Caused by: java.lang.NoClassDefFoundError: javax/ws/rs/ext/ExceptionMapper`
```

- disable modules `303-quarkus-vertx-sql` and `304-quarkus-vertx-routes` because of this native build image issue: https://github.com/quarkusio/quarkus/issues/21516

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,8 +30,9 @@
         <module>300-quarkus-vertx-webClient</module>
         <module>301-quarkus-vertx-kafka</module>
         <module>302-quarkus-vertx-jwt</module>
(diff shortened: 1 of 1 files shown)


kshpak:
@pjgg CI fails

pjgg:
Native compilation fails because only runs the modules that have been modified. In this case, we have "removed" these modules from the main pom modules, until the above issue is fixed.

kshpak (approved):

PR #271, opened 2022-01-28
Define Assertj version

gsmet:


gsmet:
Ah was doing them in batch and missed Rostislav already fixed it.

PR #278, opened 2022-04-22
Application name is now propagated in JVM tests

rsvoboda:
Application name is now propagated in JVM tests

Details in https://github.com/quarkusio/quarkus/issues/25097

I first thought there was a bug in Quarkus main, but the change goes in the right direction and we need to update the tests.


michalvavrik (approved):
Lgtm, thank you for the findings.

PR #290, opened 2022-08-10
Update 022-quarkus-properties-config scenario due to an upstream changes

pjgg:
Update test expectation in native mode


michalvavrik:
Let's wait for the first response in the issue you created. I wanna be sure this is a bug and not an intention. Please see this change (looks like completely same situation, so the author knew about this!) https://github.com/quarkusio/quarkus/pull/26802/files#diff-976daf3604137825d4254dc83dc6643ee2b1936011f47fa4023d5487083011baL43

michalvavrik (approved):
LGTM

PR #291, opened 2022-09-12
Set JDK17 as default native docker image

pjgg:
By default native images are going to be built by graalVM/mandrel JDK17

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -40,7 +40,7 @@ jobs:
     strategy:
       matrix:
         java: [ 11 ]
-        image: [ "ubi-quarkus-native-image:22.2-java11"]
+        image: [ "ubi-quarkus-native-image:22.2-java17"]
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/pr.yaml
```diff
@@ -140,7 +140,7 @@ jobs:
     strategy:
       matrix:
         java: [ 11 ]
-        graalvm-version: ["22.2.0.java11"]
+        graalvm-version: ["22.2.java17"]
```
jsmrcka, line 143:
```suggestion
        graalvm-version: ["22.2.0.java17"]
```
See the failing CI action.

I think this might also be the case in https://github.com/quarkus-qe/quarkus-test-framework/pull/561, where the CI didn't show it and I missed it too.

jsmrcka (changes requested):

rsvoboda:
pls rebase

jsmrcka (approved):
I think we're good to go.

PR #307, opened 2022-11-21
Bump used GH actions

rsvoboda:
Bump used GH actions

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -11,8 +11,8 @@ jobs:
       matrix:
         java: [ 11 ]
     steps:
-      - uses: actions/checkout@v1
-      - uses: actions/cache@v1
+      - uses: actions/checkout@v3
+      - uses: actions/cache@v3
         with:
           path: ~/.m2/repository
           key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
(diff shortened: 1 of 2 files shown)


.github/workflows/daily.yaml
```diff
@@ -21,7 +21,7 @@ jobs:
       - name: Reclaim Disk Space
         run: .github/ci-prerequisites.sh
       - name: Install JDK {{ matrix.java }}
-        uses: joschi/setup-jdk@e87a7cec853d2dd7066adf837fe12bf0f3d45e52
+        uses: joschi/setup-jdk@v2
```
afalhambra, line 24: I think we should switch to use [`actions/setup-java`](https://github.com/actions/setup-java) GHA instead, since [`joschi/setup`](https://github.com/joschi/setup-jdk/releases/tag/v2.5.1) is deprecated (also it still uses Node12 which is already deprecated by [GH](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/))

rsvoboda: oki, I can update that

rsvoboda: PR updated

.github/workflows/daily.yaml
```diff
@@ -21,8 +21,9 @@ jobs:
       - name: Reclaim Disk Space
         run: .github/ci-prerequisites.sh
       - name: Install JDK {{ matrix.java }}
-        uses: joschi/setup-jdk@e87a7cec853d2dd7066adf837fe12bf0f3d45e52
+        uses: actions/setup-java@v3
         with:
+          distribution: 'adopt-hotspot'
```
afalhambra, line 26: `Adopt OpenJDK` is moved to Eclipse Temurin, [see](https://github.com/actions/setup-java/blob/v2/README.md#supported-distributions).

So it won't be updated anymore, I think we should used `temurin` as it is also the one we have defined in our test-suite. For instance, [here](https://github.com/quarkus-qe/quarkus-test-suite/blob/main/.github/workflows/daily.yaml#L20).

rsvoboda: OK, wanted to have diversity across our TS, but will move to `temurin`, same as the main TS

PR #308, opened 2022-11-22
Update native image builder to 22.3 for Mandrel and GraalVM

afalhambra:
### Summary

`quay.io/quarkus/ubi-quarkus-mandrel:22.2-java17` -> `quay.io/quarkus/ubi-quarkus-mandrel:22.3-java17`
`quay.io/quarkus/ubi-quarkus-mandrel:22.2-java11` -> `quay.io/quarkus/ubi-quarkus-mandrel:22.3-java17`
`quay.io/quarkus/ubi-quarkus-native-image:22.2-java17` -> `quay.io/quarkus/ubi-quarkus-native-image:22.3-java17`

Please select the relevant options.
- [X] Dependency update

### Checklist:
- [X] Commits are well encapsulated and follow [the best practices](https://cbea.ms/git-commit/)

PR #309, opened 2022-11-22
Follow-up regarding GH Actions updates

afalhambra:
Follows up https://github.com/quarkus-qe/beefy-scenarios/pull/307

Updating below github actions:

From `DeLaGuardo/setup-graalvm@master` to `graalvm/setup-graalvm@v1`
From `jitterbit/get-changed-files@v1` to `tj-actions/changed-files@v34.4.3`

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -176,7 +176,8 @@ jobs:
     strategy:
       matrix:
         java: [ 11 ]
-        graalvm-version: ["22.2.0.java11"]
+        graalvm-version: [ "22.2.0" ]
+        graalvm-java-version: [ "11" ]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/cache@v3
@@ -203,9 +204,11 @@ jobs:
       - uses: microsoft/setup-msbuild@v1
       - name: Setup GraalVM
         id: setup-graalvm
-        uses: DeLaGuardo/setup-graalvm@master
+        uses: graalvm/setup-graalvm@v1
         with:
-          graalvm-version: ${{ matrix.graalvm-version }}
+          version: ${{ matrix.graalvm-version }}
+          java-version: ${{ matrix.graalvm-java-version }}
+          components: 'native-image'
       - name: Install native-image component
         run: |
           gu.cmd install native-image
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -107,7 +107,7 @@ jobs:
         shell: bash
         run: tar -xzf maven-repo.tgz -C ~
       - id: files
-        uses: jitterbit/get-changed-files@v1
+        uses: tj-actions/changed-files@v34.0.0
         continue-on-error: true
       - name: Test in Native mode
         run: |
@@ -118,7 +118,7 @@ jobs:
 
           for module in $MODULES
           do
-              if [[ "${{ steps.files.outputs.all }}" =~ ("$module") ]] ; then
+              if [[ "${{ steps.files.outputs.all_changed_and_modified_files }}" =~ ("$module") ]] ; then
                   CHANGED=$(echo $CHANGED" "$module)
               fi
           done
@@ -143,7 +143,8 @@ jobs:
     strategy:
       matrix:
         java: [ 11 ]
-        graalvm-version: ["22.2.0.java17"]
+        graalvm-version: ["22.2.0"]
+        graalvm-java-version: [ "17" ]
     steps:
       - uses: actions/checkout@v3
       - uses: actions/cache@v3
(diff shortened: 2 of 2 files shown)


afalhambra:
Blocks https://github.com/quarkus-qe/beefy-scenarios/pull/308

.github/workflows/daily.yaml
```diff
@@ -199,13 +199,14 @@ jobs:
         shell: bash
         run: tar -xzf maven-repo.tgz -C ~
       - name: Install cl.exe
-        uses: ilammy/msvc-dev-cmd@v1
```
michalvavrik, line 202: Please see https://github.com/ilammy/msvc-dev-cmd/tags, both ilammy/msvc-dev-cmd@v1 and ilammy/msvc-dev-cmd@v1.12.0 are tags for commit [7315a94](https://github.com/ilammy/msvc-dev-cmd/commit/7315a94840631165970262a99c72cfb48a65d25d). Why is this change needed?

.github/workflows/daily.yaml
```diff
@@ -199,13 +199,14 @@ jobs:
...
-      - uses: microsoft/setup-msbuild@v1
+        uses: ilammy/msvc-dev-cmd@v1.12.0
+      - uses: microsoft/setup-msbuild@v1.1.3
```
michalvavrik, line 203: Please see https://github.com/microsoft/setup-msbuild/tags, both ilammy/msvc-dev-cmd@v1 and microsoft/setup-msbuild@v1.1.3 are tags for https://github.com/microsoft/setup-msbuild/commit/34cfbaee7f672c76950673338facd8a73f637506. Why is this change needed?

.github/workflows/pr.yaml
```diff
@@ -107,7 +107,7 @@ jobs:
         shell: bash
         run: tar -xzf maven-repo.tgz -C ~
       - id: files
-        uses: jitterbit/get-changed-files@v1
+        uses: tj-actions/changed-files@v34.4.3
```
michalvavrik, line 110: Could you stick with Major?

Could you describe why is this change needed?

michalvavrik (changes requested):

.github/workflows/daily.yaml
afalhambra: same reasoning as before, but for this GHA.

.github/workflows/pr.yaml
afalhambra: There hasn't been any [activity](https://github.com/jitterbit/get-changed-files) for this GHA `jitterbit/get-change-files` for more than 1 year and a half, and it's about to get deprecated due to Node12 not being updated. So I decided to get a more updated and active GHA replacement.

.github/workflows/daily.yaml
michalvavrik: You say we should switch from https://github.com/ilammy/msvc-dev-cmd/commit/7315a94840631165970262a99c72cfb48a65d25d to https://github.com/ilammy/msvc-dev-cmd/commit/7315a94840631165970262a99c72cfb48a65d25d because v1 is same as v.12.0

.github/workflows/daily.yaml
michalvavrik: I say you are actually not changing anything, just saying we should stick with some minor version. Both tags target same commit.

.github/workflows/daily.yaml
afalhambra: sorry, not sure if I get you... 

I'm just trying to update that GHA from `v1` to `v1.12.0` as latter is the one updated to use Node16

.github/workflows/daily.yaml
michalvavrik: image

.github/workflows/daily.yaml
afalhambra: thank you @michalvavrik, didn't spot it

afalhambra: done

.github/workflows/daily.yaml
afalhambra: thank you @michalvavrik, didn't spot it.
Done

michalvavrik (commented):
Please check Windows failure: _##[error]Unexpected HTTP response: 404_

rsvoboda (approved):
We should be using GraalVM 22.3, but that can be handled in follow-up PR

michalvavrik (approved):
Thanks for changes.

afalhambra:
> We should be using GraalVM 22.3, but that can be handled in follow-up PR

@rsvoboda I already opened it this morning https://github.com/quarkus-qe/beefy-scenarios/pull/308 but I was waiting for this PR to be merged first.

So will rebase that one later today or tomorrow

PR #310, opened 2022-11-23
Remove non-used GHA for Windows Native build

afalhambra:
Follows up of https://github.com/quarkus-qe/beefy-scenarios/pull/309

As a [`graalvm/setup-graalvm`](https://github.com/graalvm/setup-graalvm) GHA was introduced, we need to get rid of the non used GHA actions for Windows, as they are now already available when using this GHA


michalvavrik (approved):
lgtm

PR #335, opened 2023-01-02
Fix changed modules detection in PRs and run only necessary CI modules

michalvavrik:
Fix based on the example https://github.com/tj-actions/changed-files#usage.


fedinskiy (approved):

michalvavrik:
Windows failure not related, but it's definitely something we might wanna address if keeps happening.

PR #337, opened 2023-01-02
Revert "Fix changed modules detection in PRs and run only necessary"

michalvavrik:
This reverts commit 90938352c7c61bd25a5c9ad41b190d476a6c9405.

Now that I understand the problem, only Quarkus QE Test Suite was affected as this PR workflow does not output changes. I'm very sorry.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -118,7 +118,7 @@ jobs:
 
           for module in $MODULES
           do
-              if [[ "${{ steps.changed-files.outputs.all_changed_and_modified_files }}" =~ ("$module") ]] ; then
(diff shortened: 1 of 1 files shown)


rsvoboda:
So we need to reference `steps.files.outputs.all_changed_and_modified_files` because we have the following definition above the `Test in Native mode` step?
```
      - id: files
        uses: tj-actions/changed-files@v35.2.0
        continue-on-error: true
```

michalvavrik:
yeah, exactly, yesterday I misunderstood example in the docs, now I understand it I know Beefy scenarios were never affected https://github.com/quarkus-qe/quarkus-test-suite/pull/972

rsvoboda (approved):

PR #357, opened 2023-02-27
Fix Test Suite compilation broken by bump to Hibernate 6.2 in Quarkus

pjgg:
- https://github.com/quarkusio/quarkus/pull/31235 upgraded Hibernate to version 6.2. org.hibernate.annotations.ParamDef#type now accepts Class instead of plain string. 

- hibernate sequence generator needs some enhances and define the increment values on the migration SQL scripts


011-quarkus-panache-rest-flyway/src/main/resources/db/migration/V1.0.0__init.sql
```diff
@@ -1,4 +1,6 @@
-CREATE SEQUENCE hibernate_sequence START 1;
+-- TODO https://github.com/quarkusio/quarkus/issues/31451
```
michalvavrik, line 1: using more than one sequence seems to be correct strategy according to the https://github.com/quarkusio/quarkus/issues/31256, so I'm not sure about the TODO

michalvavrik (changes requested):

michalvavrik (approved):

michalvavrik:
`021-quarkus-panache-multiple-pus` and `017-quartz-cluster` are failing, however only changes are in `011-quarkus-panache-rest-flyway` and that's one is a fixed, so I call it an improvement and merging. Thanks

011-quarkus-panache-rest-flyway/src/main/resources/db/migration/V1.0.0__init.sql
```diff
@@ -1,4 +1,5 @@
-CREATE SEQUENCE hibernate_sequence START 1;
+CREATE SEQUENCE application_SEQ START 1 INCREMENT 50;
```
rsvoboda, line 1: @pjgg @michalvavrik why increment by 50?

This feels almost like a bug. What's the reason for increment by 50?

michalvavrik: it is not a bug, we can't use hibernate sequence and each table needs to have it's own sequence. I didn't argue with increment (though I don't know reason) as it strongly depends on allocation policy and pool strategy

michalvavrik: What I mean is that Hibernate Schema has sequence increment 50, so we can't use 1. But we could define allocation on the entity.

rsvoboda: https://jakarta.ee/specifications/persistence/2.2/apidocs/javax/persistence/sequencegenerator#allocationSize() says Default:  50 

That's probably the case.

What's strange to me is that increment definition wasn't needed before

michalvavrik: before we weren't using our sequence, but hibernate one. increment by 50 is a good thing in real life app; f.e. here in the first comment, I found explanation for pros of 50 https://discourse.hibernate.org/t/generated-value-strategy-auto/6481

PR #366, opened 2023-03-02
Remove quarkus-opentelemetry-exporter-otlp and fix traces operation name in one module

michalvavrik:
- Quarkus OpenTelemetry now contains OTLP exported, there is no need to extra dependency
- Fix traces operation name in module `300-quarkus-vertx-webclient` after OT bump to the 1.23.1 - https://github.com/quarkusio/quarkus/pull/31356

I'm also aware there is issue with traces operation name in module 301, but I will address it in a separate PR as I'm having issue to run these tests at all (I hit serialization problems).

### DIFF
--- a/300-quarkus-vertx-webClient/pom.xml
+++ b/300-quarkus-vertx-webClient/pom.xml
@@ -14,10 +14,6 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-opentelemetry-exporter-otlp</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>io.quarkus</groupId>
             <artifactId>quarkus-vertx</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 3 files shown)


pjgg:
Remove `quarkus-opentelemetry-exporter-otlp` itś fine because is included on `quarkus-opentelemetry` by default but we have to understand why `neo4j` is not resolved

002-quarkus-all-extensions/pom.xml
```diff
@@ -140,10 +140,6 @@
             <artifactId>woodstox-core</artifactId>
             <version>6.5.0</version>
         </dependency>
-        <dependency>
```
pjgg, line 143: Which error do you have at compile time with Neo4j dep?

pjgg (changes requested):

002-quarkus-all-extensions/pom.xml
michalvavrik: It can't find 999-SNAPSHOT for neo4j locally, I think it is my local (relocation) problem. I'll drop it and just remove exported otlp.

rsvoboda: You need to have enabled relocations module as neo4j moved to quarkiverse 

e.g. `mvn clean install -Dquickly -Prelocations  -Dno-test-modules`

michalvavrik: didn't know, thank you

michalvavrik:
Hmm, now that I think of it. CI will fail anyway as I didn't handle 301 module. I'll make it draft then.

PR #122, opened 2021-03-08
Avoid ports collision in 020-quarkus-http-non-application-endpoints for native mode

pjgg:
When you are running your test, using `QuarkusProdModeTest` you should avoid using `@NativeImageTest` annotation, otherwise, you will have a collision between ports. because `@NativeImageTest` will launch an app over port 8081 and `QuarkusProdModeTest` will launch another app per test, over the same port. 

We have defined a 'quarkus.profile' system property on `maven-failsafe-plugin`. Based on this property, we will build a native image or a JVM instance.

PR #368, opened 2023-03-08
Remove the outdated property

fedinskiy:
Property %prod.quarkus.datasource.jdbc.url was used by integration tests as a workaround for connecting to H2 database.
Due to new changes, the profile we use now is `native` instead of `prod`, and this property leads to failures.

[1] https://github.com/quarkusio/quarkus/pull/31465

### DIFF
--- a/603-spring-web-smallrye-openapi/src/main/resources/application.properties
+++ b/603-spring-web-smallrye-openapi/src/main/resources/application.properties
@@ -2,7 +2,6 @@ spring.application.name = Bootstrap Spring Boot
 
 quarkus.datasource.db-kind=h2
 quarkus.datasource.jdbc.url=jdbc:h2:mem:bootapp;DB_CLOSE_DELAY=-1;DB_CLOSE_ON_EXIT=FALSE
-%prod.quarkus.datasource.jdbc.url=jdbc:h2:tcp://localhost/mem:test
 
 quarkus.hibernate-orm.database.generation=drop-and-create
 


603-spring-web-smallrye-openapi/src/main/resources/application.properties
```diff
@@ -2,7 +2,6 @@ spring.application.name = Bootstrap Spring Boot
 
 quarkus.datasource.db-kind=h2
 quarkus.datasource.jdbc.url=jdbc:h2:mem:bootapp;DB_CLOSE_DELAY=-1;DB_CLOSE_ON_EXIT=FALSE
-%prod.quarkus.datasource.jdbc.url=jdbc:h2:tcp://localhost/mem:test
```
jsmrcka, line 5: Ok, so this URL is heavily outdated anyway (since ~ Quarkus 2.2).

This means that the `quarkus.datasource.jdbc.url` works for both JVM and native, as long as the `@QuarkusTestResource(H2DatabaseTestResource.class)` is present. Is that correct?

And do you know what is the motivation for the `<quarkus.profile>native</quarkus.profile>` in the `native` maven profile?

jsmrcka (approved):

603-spring-web-smallrye-openapi/src/main/resources/application.properties
fedinskiy: Yes, ` quarkus.datasource.jdbc.url` now works for both JVM and native.

It looks like the intention was to avoid a conflict over ports: https://github.com/quarkus-qe/beefy-scenarios/pull/122

PR #404, opened 2023-08-05
Fix Quarkus Vert.x scenarios with Kafka failures

michalvavrik:
Daily build is failing for several days in row. TBH what I experienced locally was little difference than daily build logs (there it was just tracing issue IMHO), but this is what I could reproduce locally:

- Kafka operation name `stock-price send` seem to changed to `stock-price publish`
- `quarkus-opentelemetry-exporter-otlp` brought javaxs staff and build failed so I dropped it and correct URL
- then tracing didn't work so I migrated exporter tracing URL
-  then Kafka serialization problem appeared and tests failed, so I migrated it to the registry v2, together with all config properties and URLs
- but then README and docker compose files were obsolete (and didn't work) so I dropped them for I don't want to invest time for something we don't use
- I am not 100 % sure, but it seems like that test is not written correctly and as a consequence, there is occasionally error logging caused by the fact that we still try to emit messages into channel that has been cancelled and after Arc shutdown we still invoke from `io.quarkus.qe.kafka.VerticleDeployer#run` generate method so there is an error injecting been 

```
2023-08-05 16:55:58,180 DEBUG [io.qua.arc.impl] (main) ArC DI container shut down
2023-08-05 16:55:58,179 ERROR [io.qua.ver.cor.run.VertxCoreRecorder] (vert.x-eventloop-thread-4) Uncaught exception received by Vert.x [Error Occurred After Shutdown]: java.lang.RuntimeException: Error injecting org.eclipse.microprofile.reactive.messaging.Emitter<io.quarkus.qe.kafka.StockPrice> io.quarkus.qe.kafka.producers.StockPriceProducer.emitter
	at io.quarkus.qe.kafka.producers.StockPriceProducer_Bean.doCreate(Unknown Source)
	at io.quarkus.qe.kafka.producers.StockPriceProducer_Bean.create(Unknown Source)
	at io.quarkus.qe.kafka.producers.StockPriceProducer_Bean.create(Unknown Source)
	at io.quarkus.arc.impl.AbstractSharedContext.createInstanceHandle(AbstractSharedContext.java:113)
```
I am not going to play with that as it doesn't impact results. I think there might be other factors (like my changes...).

https://github.com/quarkiverse/quarkus-apicurio-registry-client is archived now, so there is no point of testing it.

Also I think the test is going to be flaky due to https://github.com/smallrye/smallrye-reactive-messaging/issues/2253, but whether we experience it again or not, I don't know. Can't reproduce it locally.


michalvavrik:
there must be some difference between my local env and CI, because I can't reproduce it. Will have a look.

michalvavrik:
Opened https://github.com/smallrye/smallrye-reactive-messaging/issues/2253

mjurc (approved):

PR #456, opened 2024-02-15
Update db2 images

jedla97:
Updating the db2 images as this is old version and it's from custom quay.io image repository.


mocenas (approved):
LGTM

PR #464, opened 2024-03-11
Move to quarkus-resteasy-client based artifacts

rsvoboda:
Move to quarkus-resteasy-client based artifacts

There is big rename going on in Quarkus main / Quarkus 3.9 Let's use the new naming and not rely on relocations which can be deleted any day

002-quarkus-all-extensions fail is unfortunately expected, it's under investigation atm.

It seems that GraalVM upgrade is triggering passing full classpath and it fails on https://github.com/quarkusio/quarkus/blob/main/core/deployment/src/main/java/io/quarkus/deployment/recording/BytecodeRecorderImpl.java#L681 for applications with big classpath, I will file issue for it once I find more details.

### DIFF
--- a/002-quarkus-all-extensions/pom.xml
+++ b/002-quarkus-all-extensions/pom.xml
@@ -54,7 +54,7 @@
         </dependency>
         <dependency>
             <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-rest-client</artifactId>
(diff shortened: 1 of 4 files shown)


rsvoboda:
So the rename happened already - https://github.com/quarkusio/quarkus/pull/39159

michalvavrik (approved):
Thanks!

PR #465, opened 2024-03-13
Move mvnw to 3.9.6

rsvoboda:
Move mvnw to 3.9.6

Triggered by https://github.com/quarkusio/quarkus/issues/39350#issuecomment-1993736747


michalvavrik (approved):
Agreed, that's right change, thanks.

PR #494, opened 2024-07-29
Update deprecated `@QuarkusTestResource` to `@WithTestResource`

jedla97:
Updating deprecated @QuarkusTestResource to @WithTestResource as it's new in 3.13.0 https://github.com/quarkusio/quarkus/wiki/Migration-Guide-3.13#quarkustestresource-replaced-by-withtestresource-gear-white_check_mark


rsvoboda (commented):
`restrictToAnnotatedClass = false` ... what's the key to add this?

I noticed you updated the PR 2 times, so the test failures triggered that?

jedla97:
Yes the update was triggered by the failures. I miss one failure with first run.

> restrictToAnnotatedClass = false ... what's the key to add this?

This is added when the test are executed outside of that class. With old annotation the default value for `restrictToAnnotatedClass` was false, bt now with `WithTestResource` the default is true.

rsvoboda (approved):

PR #498, opened 2024-08-14
Use dereference-schema=true as workaround for quarkus #42504 issue

rsvoboda:
Use dereference-schema=true as workaround for quarkus #42504 issue


michalvavrik (approved):
I still think it's a bug, but let's have CI green...

rsvoboda:
+1, the main motivation is to have green CI

PR #503, opened 2024-08-27
Use quarkus-micrometer-registry-prometheus

rsvoboda:
Use `quarkus-micrometer-registry-prometheus` instead of direct `io.micrometer:micrometer-registry-prometheus` usage


rsvoboda:
Approved by Bruno, CI is green, merging.

PR #508, opened 2024-09-09
Reduce log size in CI runs by suppressing download transfer progress

gtroitsk:
Reduces log heaviness. Adds `--batch-mode --no-transfer-progress` for maven execution to reduce "Downloading ...." messages


jedla97 (approved):
Thanks for continue this series of reducing the logs!

PR #510, opened 2024-09-20
Project cleanup 29.9.2024

gtroitsk:
- Enables disabled tests, where issue was resolved
- Adds new extensions to  002 and 003 many extensions modules
- Adds Keycloak support to 002 and 003 many extensions modules
- Migrates 013-quarkus-oidc-restlcient to Keycloak Dev Svc 

- Removes pagefile configuration from CI Windows workflows as GitHub-hosted runners double the power for open source projects


003-quarkus-many-extensions/pom.xml
```diff
@@ -201,11 +195,15 @@
...
-          <groupId>io.quarkus</groupId>
-          <artifactId>quarkus-reactive-pg-client</artifactId>
-        </dependency> -->
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-reactive-pg-client</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
```
michalvavrik, line 204: IMHO this is a wrong approach. Testcontainer and `@WithTestResource` etc. are heavily tested upstream. Users are supposed to use Dev Services or OIDC Wiremock test resource, let's use them unless there is a good reason. It's harder to maintain and will take are more time then setting couple of dev svc config properties.

101-javaee-like-getting-started/pom.xml
```diff
@@ -47,10 +46,10 @@
             <groupId>io.quarkus</groupId>
             <artifactId>quarkus-scheduler</artifactId>
         </dependency>
-        <!-- <dependency>  https://github.com/quarkusio/quarkus/issues/2725
-      <groupId>io.quarkus</groupId>
-      <artifactId>quarkus-smallrye-jwt</artifactId>
```
michalvavrik, line 52: Commit is called `Add new dependency to 101 javaee getting-started project as issue was…
… resolved` but you are adding SR JWT. I don't understand it.

303-quarkus-vertx-sql/src/test/java/io/quarkus/qe/vertx/sql/dbpool/PostgresPoolTest.java
```diff
@@ -135,47 +126,6 @@ public void checkIdleExpirationTime() throws InterruptedException {
         assertEquals(doneIdleExpired.getCount(), 0, "Missing doneIdleExpired query.");
     }
 
-    @Test
-    @DisplayName("Idle issue: Fail to read any response from the server, the underlying connection might get lost unexpectedly.")
-    @Order(3)
-    @Disabled("Takes too much time and is fixed by Vertx 4.1")
-    public void checkBorderConditionBetweenIdleAndGetConnection() {
```
michalvavrik, line 142: +1; we have PG POol tests in the TS and they are slow (and probably flaky)

002-quarkus-all-extensions/src/main/resources/application.properties
```diff
@@ -18,14 +18,12 @@ quarkus.kafka-streams.topics=kafka-topic
...
-#quarkus.oauth2.client-id=client_id
-#quarkus.oauth2.client-secret=secret
-#quarkus.oauth2.introspection-url=http://oauth-server/introspect
-
-# Disabled needs running keycloak server
-#quarkus.oidc.auth-server-url=https://localhost:8180/auth/realms/quarkus
-#quarkus.oidc.client-id=quarkus-app
+quarkus.oauth2.client-id=client_id
+quarkus.oauth2.client-secret=secret
+quarkus.oauth2.introspection-url=http://oauth-server/introspect
```
michalvavrik, line 23: I have doubt that these 3 properties has any effect at all. Do they?

002-quarkus-all-extensions/pom.xml
```diff
@@ -104,15 +103,14 @@
...
-          <groupId>io.quarkus</groupId>
-          <artifactId>quarkus-hibernate-search-orm-elasticsearch</artifactId>
-        </dependency> -->
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-keycloak-authorization</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-oidc</artifactId>
```
michalvavrik, line 112: You do enable these 2 extensions (oidc, kc authz) and add `KeycloakTestResource` for them. However you don't have any tests calling them. The whole point I take it is that they are present and enabled. That is OK, but I think it's bit overkill to add dedicated Keycloak resource, start it when we don't use it. You set:

```
quarkus.oidc.auth-server-url=http://localhost:8180/auth/realms/test-realm
quarkus.oidc.client-id=quarkus-app
```

I think it would test about same because OIDC would provide auth server itself if you didn't set these properties.

Anyway, that is just an opinion, you can keep it. You can say it's healthy to test dedicated resource server instead of Dev Svc.

002-quarkus-all-extensions/pom.xml
```diff
@@ -336,15 +297,10 @@
...
-        <!--        <artifactId>quarkus-logging-sentry</artifactId>-->
-        <!--      </dependency>-->
-        <dependency>
-            <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-azure-functions-http</artifactId>
-        </dependency>
+        <!--        <dependency>-->
+        <!--            <groupId>io.quarkus</groupId>-->
+        <!--            <artifactId>quarkus-azure-functions-http</artifactId>-->
+        <!--        </dependency>-->
```
michalvavrik, line 303: This one I don't understand. You are disabling it? If so, why? It's hard to track what is happening here.

michalvavrik (commented):
LGTM

002-quarkus-all-extensions/src/main/resources/application.properties
gtroitsk: Yes, they are required by `quarkus-elytron-security-oauth2` extension. 
https://quarkus.io/guides/security-oauth2#setting-up-application-properties

002-quarkus-all-extensions/pom.xml
gtroitsk: There is a conflict with other extensions, I will add comment

002-quarkus-all-extensions/pom.xml
gtroitsk: Oidc doesn't run server by itself and will require set properties. I want to leave test with resource. It will cost our CI nothing as It is not running here because it requires a lot of RAM.

101-javaee-like-getting-started/pom.xml
gtroitsk: I renamed commit to `Add quarkus-smallrye-jwt extension to 101 javaee getting-started project as issue was resolved`

002-quarkus-all-extensions/src/main/resources/application.properties
michalvavrik: sorry, that's not the answer, `quarkus.oauth2.introspection-url=http://oauth-server/introspect` is invalid, so what that it is required as long as it is not as much as validated or used at all?

002-quarkus-all-extensions/pom.xml
michalvavrik: > Oidc doesn't run server by itself and will require set properties. 

not true


>  I want to leave test with resource. 

that's ok

> It will cost our CI nothing as It is not running here because it requires a lot of RAM.

I don't get it. How starting a container cost nothing and if it is not running, why are you adding it?

002-quarkus-all-extensions/src/main/resources/application.properties
michalvavrik: Anyway, I'll close this comment as this extension is clearly not used at all. There is no point of talking about it.

002-quarkus-all-extensions/pom.xml
```diff
@@ -470,32 +397,31 @@
...
         <dependency>
             <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-picocli</artifactId>
-        </dependency> -->
-        <dependency>
-            <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-reactive-db2-client</artifactId>
+            <artifactId>quarkus-minikube</artifactId>
         </dependency>
+        <!--  We cannot have picocli and REST assured at the same time. -->
```
michalvavrik, line 420: that's hard to believe, why?

002-quarkus-all-extensions/src/main/resources/application.properties
gtroitsk: That's the part of minimal config according to docs and it caused exception `Caused by: io.quarkus.runtime.configuration.ConfigurationException: client-id, client-secret and introspection-url must be configured when the oauth2 extension is enabled
`

002-quarkus-all-extensions/pom.xml
```diff
@@ -336,15 +297,11 @@
...
-        <!--        Caused by: java.lang.RuntimeException: java.lang.ClassCastException: io.sentry.servlet.SentryServletContainerInitializer cannot be cast to javax.servlet.ServletContainerInitializer-->
-        <!--      <dependency>-->
-        <!--        <groupId>io.quarkus</groupId>-->
-        <!--        <artifactId>quarkus-logging-sentry</artifactId>-->
-        <!--      </dependency>-->
-        <dependency>
-            <groupId>io.quarkus</groupId>
-            <artifactId>quarkus-azure-functions-http</artifactId>
-        </dependency>
+        <!--        Disabled because of: Multiple producers of item class io.quarkus.vertx.http.deployment.RequireVirtualHttpBuildItem-->
```
michalvavrik, line 300: sounds like a bug, which extension does it conflict it with?

002-quarkus-all-extensions/pom.xml
gtroitsk: We don't run 002 module in CI (reason above), you can run this module locally

002-quarkus-all-extensions/src/main/resources/application.properties
michalvavrik: > That's the part of minimal config according to docs and it caused exception `Caused by: io.quarkus.runtime.configuration.ConfigurationException: client-id, client-secret and introspection-url must be configured when the oauth2 extension is enabled `

I am not sure how to state it differently, the introspection URL is clearly invalid, hence it is clearly not used or validated. I don't care about required config property. The point here is, it brings no value to use this apart of having this extension present but not used.

Let's not continue.

002-quarkus-all-extensions/pom.xml
michalvavrik: :-D

michalvavrik: Who will ever do that? Let's drop that module.

002-quarkus-all-extensions/src/main/resources/application.properties
michalvavrik: Of course, keep the property. I was just commenting on the point that extension is not used.

002-quarkus-all-extensions/pom.xml
michalvavrik: Anyway, in that case I don't care, I'll close this comment.

002-quarkus-all-extensions/pom.xml
michalvavrik: FYI we do use REST Assured with Picocli in the QE TS.

michalvavrik: Probably result of missing Picocli command. That sounds valid, let's drop it or add a command.

002-quarkus-all-extensions/pom.xml
gtroitsk: Removed because of conflict with amazon http lambda extension

michalvavrik (approved):
LGTM!

PR #519, opened 2024-11-07
Fix and refactor the scope of some resources

jedla97:
There was a bug in `TestResourceScope.MATCHING_RESOURCES` which made this work and was fixed in https://github.com/quarkusio/quarkus/pull/44279, I change it in this PR https://github.com/quarkus-qe/beefy-scenarios/pull/505 but it seems I wrongy understand the docs back then.

Not it should work as:
* `RESTRICTED_TO_CLASS` start for every test class, every test class need anotation
* `MATCHING_RESOURCES` start for specific group of test classes, every test class need anotation
* `GLOBAL` start for every test class, only one annotation

I change it to cover all 3 usages.

### DIFF
--- a/.github/workflows/daily.yaml
+++ b/.github/workflows/daily.yaml
@@ -116,7 +116,7 @@ jobs:
       - name: Test in Native mode
         run: |
           mvn -fae -V -B --no-transfer-progress -s .github/mvn-settings.xml -fae clean verify -Dnative \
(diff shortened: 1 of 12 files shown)


rsvoboda:
Pleas add commit similar to https://github.com/quarkus-qe/beefy-scenarios/commit/c48ec1d1423aab1ca609a314719fbc003dde95fb to increase the memory to 8g

rsvoboda (approved):

PR #538, opened 2025-03-27
Docker/Podman is not needed for 003-quarkus-many-extensions

rsvoboda:
* Docker/Podman is not needed for 003-quarkus-many-extensions

Squeezed in:
* Drop quarkus-plugin.version property, upstream doesn't use it for ages


fedinskiy (commented):
I see changes related to removal of separate version of quarkus-plugin. I will not ask for a separate commit, but an explanation would be nice to have.

rsvoboda:
That was miss on my side, PR updated

fedinskiy (approved):

PR #592, opened 2026-04-13
Drop ubi-quarkus-mandrel-builder-image:jdk-21 usage

rsvoboda:
Drop ubi-quarkus-mandrel-builder-image:jdk-21 usage

ubi-quarkus-mandrel-builder-image:jdk-25 is the way to go, one supported Mandrel version


jedla97 (approved):

rsvoboda:
`301-quarkus-vertx-kafka` failure is another issue, I will address that soon.

PR #595, opened 2026-04-16
Test order in AgroalPoolTest

rsvoboda:
Test order in AgroalPoolTest

This should avoid `AgroalPoolTest.idleTimeoutTest:70 agroalCheckIdleTimeout: Expected 1 active connections ==> expected: <1> but was: <3>`


jcarranzan:
Ok, let's see CI checks.
If the test method order is not enough to fix it, we could also apply the changes done in our ts here https://github.com/quarkus-qe/quarkus-test-suite/pull/2205/changes

jcarranzan (approved):

PR #605, opened 2026-06-30
Build and use Quarkus 3.999-SNAPSHOT instead of 999-SNAPSHOT

jedla97:
Build and use Quarkus 3.999-SNAPSHOT instead of 999-SNAPSHOT. For some time we need to use Quarkus 3.x branch as it used for 3.x streams and the main is now for Quarkus 4.x

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -32,7 +32,7 @@ jobs:
           java-version: ${{ matrix.java }}
       - name: Build Quarkus main
         run: |
-          git clone https://github.com/quarkusio/quarkus.git && cd quarkus && ./mvnw -B --no-transfer-progress -s .github/mvn-settings.xml clean install -Dquickly -Prelocations -Dno-test-modules
+          git clone -b 3.x https://github.com/quarkusio/quarkus.git && cd quarkus && ./mvnw -B --no-transfer-progress -s .github/mvn-settings.xml clean install -Dquickly -Prelocations -Dno-test-modules -Dscalpel.enabled=false
       - name: Tar Maven Repo
         shell: bash
         run: tar -I 'pigz -9' -cf maven-repo.tgz -C ~ .m2/repository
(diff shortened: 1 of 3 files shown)


.github/workflows/pr.yaml
```diff
@@ -31,11 +31,24 @@ jobs:
...
-          git clone https://github.com/quarkusio/quarkus.git && cd quarkus && ./mvnw -B --no-transfer-progress -s .github/mvn-settings.xml clean install -Dquickly -Prelocations -Dno-test-modules
+          git clone -b 3.x https://github.com/quarkusio/quarkus.git && cd quarkus && ./mvnw -B --no-transfer-progress -s $GITHUB_WORKSPACE/.github/mvn-settings.xml clean install -Dquickly -Prelocations -Dno-test-modules -Dscalpel.enabled=false
+      - name: Tar Maven Repo
+        continue-on-error: true
+        shell: bash
+        run: cat /home/runner/.m2/repository/io/quarkus/quarkus-project/3.999-SNAPSHOT/maven-metadata.xml
+      - name: Tar Maven Repo
+        continue-on-error: true
+        shell: bash
+        run: cat cat /home/runner/.m2/repository/io/quarkus/quarkus-project/3.999-SNAPSHOT/quarkus-project-3.999-SNAPSHOT.pom
```
rsvoboda: `cat cat` ?

jedla97: I was testing something and probably copy the path with cat.

jedla97:
Ok I think I figured why it failed. The problem was that the scalpel (replacement of GIB) detect that I changed pom and decide to start for quarkus build with not all modules but only affected modules. Afaik I think if the switch didn't happen we would have hit this in future PR anyway

I added `-Dscalpel.enabled=false` when building on PR as this won't happen when building daily as there is no changes.

Other option would be use cache as we doing in FW and TS, or use Quarkus action for snapshots. I think it's not problem to build and zip the maven repository in this repo CI as the PR here are pretty rare, so build is not that problematic.

rsvoboda (approved):

rsvoboda:
> I added -Dscalpel.enabled=false when building on PR as this won't happen when building daily as there is no changes.

Please mention this in qe room @jedla97

PR #616, opened 2026-09-16
Update Java version from 17 to 21 in workflow

gastaldi:


michalvavrik (approved):
Not sure if 25 is not desirable, but definitely better, thanks!

.github/workflows/quarkus-snapshot.yaml
michalvavrik: On unrelated note: this line is seriously out of date :-)

gastaldi:
The other workflows use 21, so I went with this for the mean time ;)

rsvoboda:
+1 for 21

